PR #15, opened 2019-05-02
Create GraalVM native image for operator

mdonkers:
This PR is aimed at creating a "native image" with GraalVM. Some additional changes are made as well (see below) as preparation for this.
Building of the native image is setup in such a way that it not conflicts with the current (OpenJDK) build image, only the resulting name is the same. Use the command `mvn package -Pnative` to build first the native image inside a Docker container, then take the program file and Dockerize it. Docker is needed locally for the building.

Other smaller changes:
- Extract to properties and upgrade dependencies + plugins to latest versions
- Fix description of configmap defaults


pom.xml
```diff
@@ -1,46 +1,22 @@
...
   <description>Kubernetes Operator for the Instana Agent</description>
 
   <properties>
-    <!-- Project properties -->
-    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <maven.compiler.target>1.8</maven.compiler.target>
+    <surefire-plugin.version>2.22.0</surefire-plugin.version>
+    <quarkus.version>0.18.0</quarkus.version>
+    <!-- quarkus.version>999-SNAPSHOT</quarkus.version -->
```
mdonkers: Should remove this commented line

fstab: done

PR #20, opened 2019-10-04
Beta

wiggzz:
## Why

Our operator is stable enough to promote it to BETA. These are a few necessary changes for getting it ready to do that.

Namely, at present the instructions for running the operator are not sufficient for OpenShift and there is some missing resource permissions on OpenShift.

## What

- Change the configuration parameter for setting the leader pod slightly (see corresponding sensor 
change: https://github.com/instana/sensors/pull/1950 )
- Update some of the documentation, specifically around configuration on openshift
- Add deploymentconfigs permissions because these are missing.
- Allow mounting host repository for development of sensors
- Add cluster name environment variable
- Put download key in a secret

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ Instana Agent Operator
 
 [![CircleCI](https://circleci.com/gh/instana/instana-agent-operator.svg?style=svg)](https://circleci.com/gh/instana/instana-agent-operator)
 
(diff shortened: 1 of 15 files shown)


wiggzz:
I have tested this on OpenShift 3.11 and it now works. @fstab can you walk me through creating a new release of this next week?

PR #22, opened 2019-10-24
Multiple leader election

wiggzz:
# Why

This fixes the current leader election mechanism which is broken because it relies on the state of the configuration being saved in the agents (which it is not if the agent crashes, restarts, or the configuration module is reloaded).

Since I already had the code here for doing a poll loop and reconciling the current state with desired state, I thought it would be best to bring that in versus reimplementing it.

This also allows, in the future, multiple leader election.

# What

- Implements a poll loop for the resources which the agent would like to lead (which will be enabled by a sensor change).
- Determines across all the pods what resources are present, and then assigns them randomly across the pods that are present.
- Handles failure cases and ensures that if a pod dies or stops wanting to lead, that it another leader is assigned.


wiggzz:
We should wait until Kubernetes Sensor 1.2.88 is released before merging this.

In the meantime, using `instana/instana-agent-operator:multiple-leader-election` as the operator image will allow testing this against our beta

PR #27, opened 2019-11-25
Fix watch crashloop related to invalid resource version

nfisher:
### Why

The Operator appears to crash when the local k8s watch resource version is too far behind the current active version. This PR aims to resolve the issue by catching the exception and re-establishing the watch.

### What

TBD

### Related

* [#4421](https://instana.kanbanize.com/ctrl_board/36/cards/4421/details/)


src/test/java/com/instana/operator/file/ClasspathFile.java
```diff
@@ -1,14 +1,14 @@
...
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.stream.Collectors;
 
-public class FileUtil {
-  public static String readFromClasspath(String resource) throws IOException {
-    try (InputStream input = FileUtil.class.getResourceAsStream(resource)) {
+public class ClasspathFile {
```
jbrisbin: Maybe this should be `ClasspathFileReader`?

src/main/java/com/instana/operator/string/Contents.java
```diff
@@ -1,6 +1,6 @@
-package com.instana.operator.util;
+package com.instana.operator.string;
 
-public class StringUtils {
+public class Contents {
```
jbrisbin: 🤔

src/main/java/com/instana/operator/resource/KubernetesResource.java
```diff
@@ -7,7 +7,7 @@
 
 import java.util.function.Predicate;
 
-public class ResourceUtils {
+public class KubernetesResource {
```
jbrisbin: I think `Util` or `Helper` has a place for utility methods that are unattached to a particular class. This seems to indicate it declares a "thing" that is a `KubernetesResource`, not helper methods to work them. JMO.

pom.xml
```diff
@@ -12,7 +12,7 @@
   <properties>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <surefire-plugin.version>2.22.0</surefire-plugin.version>
-    <quarkus.version>0.19.1</quarkus.version>
+    <quarkus.version>1.0.0.Final</quarkus.version>
```
jbrisbin: ❤️

src/main/java/com/instana/operator/resource/KubernetesResource.java
nfisher: I'm not particularly a fan of the `util` as it becomes a dumping ground. This class could be split into pod and metadata.

pom.xml
nfisher: It built and passed so that means it's ok am-i-right?

jbrisbin: #worksforme ?

Still needs Susie et al to do some verification testing to ensure we don't have regressions. I think it should be fine, but we should be sure in case...

src/main/java/com/instana/operator/resource/KubernetesResource.java
jbrisbin: It could be if left unchecked and used indiscriminately. But the alternative is to not indicate it's just a holder class for helper methods. Not sure that's a net gain IMO.

jbrisbin: Might be as simple as using the plural form rather than singular. `KubernetesResources` seems better to me than just the singular form, which implies something else in my mind.

src/main/java/com/instana/operator/string/Contents.java
jbrisbin: `StringContents`? I know you're trying to get away from Utils but just plain `Contents` doesn't seem to have enough gravitas... ;)

nfisher: All of the methods accept CharSequence's so named it CharSequenceContent. Not in love with it but 🤷‍♂

src/main/java/com/instana/operator/resource/KubernetesResource.java
nfisher: Will let it ruminate in the back of my head for now.

wiggzz:
@nfisher do you want to keep this branch?

dlbock:
Closed in favor of this: https://github.com/instana/instana-agent-operator/pull/32

PR #29, opened 2020-02-13
Liveness Checks and expose Environment Variables

wiggzz:
# Why

We changed the liveness checks in the other configurations (helm, yaml) and also introduced the ability to define arbitrary environment variables in the helm chart.

# What

- Bring the liveness checks in line with the other configurations
- Expose the ability to set arbitrary environment variables via the custom resource. This makes a lot of the values in the Custom Resource redundant as all they do is set an environment variable on the agent - we shouldn't need to maintain this here.
- Remove default `INSTANA_AGENT_HTTP_LISTEN=*` since this is not the default for other configurations.
- Add a test for the `AgentDeployer#newDaemonSet` method which (at this point) only checks the new environment variable logic.
- Update `customresource-max.yaml` with possible values of some environment variables, although this list is not exhaustive and probably doesn't need to be here.


src/test/java/com/instana/operator/customresource/InstanaAgentSpecDeserializeTest.java
```diff
@@ -55,21 +55,14 @@ void testDefaultValues() throws IOException {
     assertThat(spec.isAgentRbacCreate(), equalTo(Boolean.parseBoolean(DEFAULT_AGENT_RBAC_CREATE)));
```
nfisher: Is there a cleaner way we can do these assertions (e.g. a spec matcher)? Not a blocker.

wiggzz: Maybe?

PR #30, opened 2020-02-27
Parameterize Image

wiggzz:
# Why

Some of our partners require that the image the operator deploys is parameterizable.

# What

This adds the ability to specify the agent image via the `agent.image` parameter in the Instana Agent custom resource.

This also improves the documentation around the parameters which are available for the operator.


wiggzz:
> Does this also mean we need to publish the instana-agent-operator to the IBM marketplace?

I hope not.

Also, apparently github allows me to edit your comments? Super weird...

wiggzz:
> LGTM. The story references parameterizing the instana-agent-operator image as well but I think that's a typo.

I guess it's already parameterizable because you would just edit the deployment yaml.

PR #32, opened 2020-04-03
Fix too old resource version error

wiggzz:
Fixes a long standing crash that occurs regularly due to a watch closing with a `too old resource version` error.

Also clean up some things in the yaml and Dockerfile.

### DIFF
--- a/src/main/java/com/instana/operator/cache/Retry.java
+++ b/src/main/java/com/instana/operator/cache/Retry.java
@@ -10,7 +10,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 public class Retry {
-  public static final Duration INITIAL_DELAY = Duration.ofSeconds(1);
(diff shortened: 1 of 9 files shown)


src/main/java/com/instana/operator/cache/Retry.java
```diff
@@ -0,0 +1,58 @@
...
+import io.reactivex.Scheduler;
+import io.reactivex.disposables.Disposable;
+import io.reactivex.disposables.Disposables;
+
+import java.time.Duration;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class Retry {
```
jbrisbin, line 12: Wish we could use native Quarkus or Quartz here rather than rolling our own. But it probably doesn't make a lot of sense just for this. But a generic retry/recover pattern that integrates with Quarkus would be nice.

wiggzz: If you know of one, would prefer to use that also! I didn't find one after a quick search though

PR #36, opened 2020-05-22
Add StatefulSet, DaemonSet, CronJobs to agent permissions

wiggzz:
They are added to the operator permissions, but unfortunately we need to add them to two places in the operator

1) for the operator to get the permissions, so it can grant them to the agent
2) for the agent itself

We should consider generating these permissions as well at some point.


dlbock:
sigh. missed these.

PR #38, opened 2020-08-17
Add PVC support

jbrisbin:


wiggzz:
Might need it here as well https://github.com/instana/instana-agent-operator/blob/develop/olm/operator-resources/instana-agent-operator.yaml

PR #43, opened 2020-10-15
Update instana-agent.daemonset.yaml

### LINKED ISSUE #42: daemonset dnspolicy should be explicitly set to ClusterFirstWithHostNet
When trying to scrape prometheus metrics from local services agent can't resolve local kubernetes service names with given url. That's because daemonset is using hostNetwork and has default dnfpolicy set to ClusterFirst. According to kubernetes documentation it should be set to ClusterFirstWithHostNet

forgems:
Currently agent can't resolve internal cluster hostnames because it has wrong (default) dns policy. Because of that it can't scrape metrics from cluster services. According to kubernetes documentation(https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy) dnsPolicy for pods with `hostNetwork: true` should be set to ClusterFirstHostNet
Fixes #42

### DIFF
--- a/src/main/resources/instana-agent.daemonset.yaml
+++ b/src/main/resources/instana-agent.daemonset.yaml
@@ -18,6 +18,7 @@ spec:
         app: instana-agent
     spec:
       serviceAccountName: instana-agent
+      dnsPolicy: ClusterFirstWithHostNet
       hostIPC: true
       hostNetwork: true
       hostPID: true


wiggzz:
Thanks for the PR. I am not sure if this will fix the issue you are having with prometheus metrics because in Kubernetes the Instana prometheus sensor will look for endpoints using a relative path - so if you specify a full path, it will be invalid. Instead you should just use a relative path like `/metrics`. Feel free to open a ticket with us if you have further questions about getting that working.

Anyway, we should definitely be using dnsPolicy: ClusterFirstWithHostNet so thanks for raising that. I'll release a new version of the operator shortly.

PR #44, opened 2020-11-17
Add labels to the Agent resources to help identify how the Agent is installed

dlbock:
### Why

As part of the hand-off of the K8s artifacts to Team Agent, one of the things we want to understand better is how our customers are installing the Instana Agent on their Kubernetes clusters.

Agents installed by Helm similarly have `app.kubernetes.io/managed-by` and `app.kubernetes.io/version` labels on the Kubernetes resources that get created (specifically the DaemonSet). We want to use these labels to build telemetry data to better understand how our customers are installing the agent.

### What

- Add `app.kubernetes.io/managed-by` and `app.kubernetes.io/version` labels to the Agent resources to help identify how the Agent is installed. The `app.kubernetes.io/managed-by` label can be added in the specific resource's YAML, but the `app.kubernetes.io/version` label has to be added dynamically from the `CustomResourceDefinition` just after the resource is loaded from its YAML. In order to test this, I had to add the `quarkus-test-kubernetes-client` dependency to get access to fabric8's Kubernetes test framework.

- The `apiextensions.k8s.io/v1beta1` version of CustomResourceDefinition is deprecated and will no longer be served in v1.19. Use `apiextensions.k8s.io/v1` instead. This is according to the ['Deprecations and Removals' section](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#deprecations-and-removals) of the Kubernetes changelog for 1.16

### DIFF
--- a/src/test/java/com/instana/operator/AgentDeployerTest.java
+++ b/src/test/java/com/instana/operator/AgentDeployerTest.java
@@ -1,31 +1,56 @@
 package com.instana.operator;
 
+import static com.instana.operator.env.Environment.RELATED_IMAGE_INSTANA_AGENT;
+import static java.net.HttpURLConnection.HTTP_OK;
+import static java.util.Collections.emptyMap;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.hasEntry;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.hasKey;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+
+import java.util.Collections;
+import java.util.Map;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
 import com.google.common.collect.ImmutableMap;
 import com.instana.operator.customresource.InstanaAgent;
 import com.instana.operator.customresource.InstanaAgentSpec;
 import com.instana.operator.env.Environment;
+
 import io.fabric8.kubernetes.api.model.Container;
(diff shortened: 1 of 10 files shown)


src/test/java/com/instana/operator/AgentDeployerTest.java
```diff
@@ -70,20 +78,75 @@ void daemonset_must_include_specified_image() {
...
   }
 
+  @Test
+  public void daemonset_must_include_version_label_if_specified_on_crd() {
+    // create crd from yaml
+    client.load(getClass().getResourceAsStream("/customresourcedefinition-with-version.yaml"))
+        .createOrReplace()
+        .stream()
+        .findAny()
+        .orElse(null);
```
wiggzz, line 103: Does this rely on having some kube context present? I don't the rest of these tests relied on this, which may be why they are failing on jenkins / circleci.

src/test/java/com/instana/operator/AgentDeployerTest.java
```diff
@@ -70,20 +78,75 @@ void daemonset_must_include_specified_image() {
...
+    // create crd from yaml and remove version label
+    client
+        .load(getClass().getResourceAsStream("/customresourcedefinition-without-version.yaml"))
+        .createOrReplace()
+        .stream()
+        .findAny()
+        .orElse(null)
+        .getMetadata()
+        .getLabels()
+        .remove("app.kubernetes.io/version");
```
wiggzz, line 132: Do you need to remove the version label if it is already not in the yaml? since this is `-without-version.yaml`?

src/test/java/com/instana/operator/AgentDeployerTest.java
dlbock: Yeah. My mistake. Fixing now.

wiggzz (approved):

PR #46, opened 2020-12-04
[Breaking change] Upgrade quarkus from 1.2.1.Final to 1.9.0.Final

dlbock:
### Why

The version of the `quarkus` dependency is sorely outdated (`1.2.1.Final`). In order for us to leverage the `v1` api of `CustomResourceDefinition` (which we will need to do soon since `v1beta1/CustomResourceDefinition` is deprecated and will be removed in Kubernetes v1.22+), we need at least `kubernetes-client 4.11.x`. Version `1.9.0.Final` of quarkus gives us `kubernetes-client 4.12`. The last version update of `kubernetes-client` Quarkus did was to `4.10.3` before that.

**Note**: This is a breaking change as the version of `kubernetes-client` that's used when calling `.createOrReplace` on the Custom Resource's status requires an additional `create` permission in the ClusterRole for `instana.io`.

### What

- The main change here is the version bump of quarkus from 1.2.1.Final to 1.9.0.Final and all the necessary code changes for it to compile. 
- The other change to point out is that we had to add the `create` permissions to the `instana.io` apiGroup in the Operator's `ClusterRole` in `instana-agent-operator.yaml`. This is due to a change in `[kubernetes-client](https://github.com/fabric8io/kubernetes-client/issues/2292)` when calling `createOrReplace`, it will first try a `create` and if it fails with a conflict exception, then it'll try an `update`, which means that the operator will need `create` permissions in the `instana.io` apiGroup
- Added some additional error messaging to inform the user that they should check the operator has updated cluster role permissions if they encounter an error when applying the custom resource
- Also updated the `operator-artifacts.jsonnet` template to include the version of the `instana/instana-agent-operator` image tag in the generated `instana-agent-operator.yaml` to help ensure that the operators using a previous tag would still work if we introduce a breaking change in a new operator image that requires an updated `instana-agent-operator.yaml` (like the one we're doing now)
- No longer pushing the `latest` tag for the operator image in order to tie the operator yaml updates more closely with the operator image updates and avoid the situation where the customers are pulling the latest operator image containing a breaking change but not explicitly updating the operator yaml

The auxiliary changes are for easing the end-to-end testing with Kind with the introduction of 2 scripts that will:
1. Create a local Kind cluster with 1 control plane and 2 nodes (with support for macOS/Darwin and Linux), build the operator Docker image with the new maven artifacts and load that Docker image into the Kind cluster, pull the latest agent Docker image and load that into the Kind cluster
2. Delete that local Kind cluster

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -33,9 +33,7 @@ pipeline {
           if (isFullRelease(TAG)) {
             docker.withRegistry('https://index.docker.io/v1/', '8a04e3ab-c6db-44af-8198-1beb391c98d2') {
               def image = docker.build("instana/instana-agent-operator:$VERSION", BUILD_ARGS)
-
               image.push()
-              image.push('latest')
             }
           } else {
             echo "Skipping pushing tag because this is a pre-release or branch."
@@ -46,7 +44,6 @@ pipeline {
               // annoyingly no way to reuse the existing image with docker jenkins plugin.
               // probably should just pull all of this into a shell script
               def image = docker.build("scan.connect.redhat.com/ospid-6da7e6aa-00e1-4355-9c15-21d63fb091b6/instana-agent-operator:$VERSION", BUILD_ARGS)
-
               image.push()
             }
           }
--- a/olm/operator-resources/instana-agent-operator.yaml
+++ b/olm/operator-resources/instana-agent-operator.yaml
@@ -67,9 +67,9 @@ rules:
   - 'events'
   verbs:
   - 'create'
-# -------------------------------------------------------------------------
-# For the custom resource, the operator needs list, watch, get, update.
-# -------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
+# For the custom resource, the operator needs list, watch, get, update, create.
+# -----------------------------------------------------------------------------
 - apiGroups:
   - 'instana.io'
   resources:
@@ -79,6 +79,7 @@ rules:
   - 'update'
   - 'list'
   - 'watch'
+  - 'create'
 # -------------------------------------------------------------------------
 # Below are the permissions are for the agent.
 # The operator needs these permissions to create the agent's cluster role.
--- a/olm/operator-resources/operator-artifacts.jsonnet
+++ b/olm/operator-resources/operator-artifacts.jsonnet
@@ -10,9 +10,15 @@ local operatorResourcesWithVersion = std.map(addVersionToMetadataLabels, operato
 
 local addVersionToDeploymentSpec(deployment) = deployment + {
   spec+: {
-    template+: { metadata+: { labels+:
-      super.labels + { "app.kubernetes.io/version": version }
-    }}
+    template+: {
+      metadata+: {
+        labels+: super.labels + { "app.kubernetes.io/version": version }
+      },
+      spec+:
+        super.spec + {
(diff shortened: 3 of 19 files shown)


olm/operator-resources/instana-agent-operator.yaml
```diff
@@ -79,6 +79,7 @@ rules:
   - 'update'
   - 'list'
   - 'watch'
+  - 'create'
```
wiggzz, line 82: I think this may cause problems if the new operator image gets re-pulled by existing operator installations and requires an update.

Would be nice if the error message that we show in the operator if we fail on an API call said something like "Please check that your operator has the cluster role permissions that match this operator version".

Actually, ideally we would be a bit more careful with our versioning e.g. we would hard code the major version of the operator in the yaml and push major version tags as well as `latest`, so that we can avoid re-pulling the image causing permissions issues.

dlbock: Added some changes to address this @wiggzz. Please re-review ♻️ when you get a chance 🙏 ❤️ !

olm/operator-resources/operator-artifacts.jsonnet
```diff
@@ -10,9 +10,15 @@ local operatorResourcesWithVersion = std.map(addVersionToMetadataLabels, operato
...
-    template+: { metadata+: { labels+:
-      super.labels + { "app.kubernetes.io/version": version }
-    }}
+    template+: {
+      metadata+: {
+        labels+: super.labels + { "app.kubernetes.io/version": version }
+      },
+      spec+:
+        super.spec + {
+          containers: std.mapWithIndex(function(i, c) if i == 0 then c + {image: c.image + ":" + version} else c, super.containers)
```
wiggzz, line 19: Is this the full version `0.3.9`? This is good for now - but ideally in the future we should probably create major version tags, (so, in this case `0`) and then use the major version tag here as opposed to the full version. But honestly we probably need a proper update strategy other than just "wait for the operator to crash and re-pull"... so, this is good.

wiggzz (commented):
Nice. Looks good.

wiggzz (approved):

olm/operator-resources/operator-artifacts.jsonnet
dlbock: Yeah this is the full version: `image: instana/instana-agent-operator:0.3.9`

Jenkinsfile
```diff
@@ -33,9 +33,7 @@ pipeline {
           if (isFullRelease(TAG)) {
             docker.withRegistry('https://index.docker.io/v1/', '8a04e3ab-c6db-44af-8198-1beb391c98d2') {
               def image = docker.build("instana/instana-agent-operator:$VERSION", BUILD_ARGS)
-
               image.push()
-              image.push('latest')
```
dlbock, line 38: @wiggzz After discussion with Henning/Miel/Et al., we decided to no longer push the `latest` tag and to make the dependency between the operator image and the operator yaml more explicit, i.e. if a customer wants a more recent version of the operator, they have to update their operator yaml.

dlbock:
Tested the image change with and without the yaml change. Also ensured that the OLM and RedHat artifacts generated only contain the expected changes.

dlbock:
Will release v1.0.0 closer to Release 192

PR #49, opened 2021-02-25
Instana-Ticket-18030 + 18127 Add Parameter to change ImagePullPolicy to IfNotPresent + OpenTelemetry support for Port in DS

asafarian:
This Pull Request is for instana support Ticket: 18030.
Currently, the Instana Daemon set always pulls an Image. In cases where you only change resources, like config which need to recreate the pod, we do not want on all our hosts that the agent is pulled again from the repo.
Therefore we would like to have this param to change the policy to only Pull If not present on node.


src/main/java/com/instana/operator/AgentDeployer.java
```diff
@@ -71,7 +49,7 @@
 import static com.instana.operator.util.ResourceUtils.hasOwner;
 import static com.instana.operator.util.ResourceUtils.hasSameName;
 import static com.instana.operator.util.ResourceUtils.name;
-import static com.instana.operator.util.StringUtils.isBlank;
+import static com.instana.operator.util.StringUtils.*;
```
mdonkers, line 52: We prefer not to use 'star' imports. For bigger lists we have less of an issue with that, but for one or two imports please reference these directly.

src/main/java/com/instana/operator/customresource/InstanaAgentSpec.java
```diff
@@ -237,6 +246,20 @@ public String getClusterName() {
...
+
+  public void setAgentOtelActive(Boolean agentOtelActive) { this.agentOtelActive = agentOtelActive; }
+
+  public Integer getAgentOtelPort() { return agentOtelPort; }
+
+  public void setAgentOtelPort(Integer agentOtelPort) {
+    this.agentOtelPort = agentOtelPort;
+  }
+
+  public void setAgentImagePullPolicy(String imagePullPolicy) { this.agentImagePullPolicy = imagePullPolicy; }
```
mdonkers, line 261: Would be nice to relocate this next to the `getAgentImagePullPolicy`.

src/main/java/com/instana/operator/AgentDeployer.java
```diff
@@ -356,6 +347,23 @@ DaemonSet newDaemonSet(InstanaAgent owner,
...
+
+    Boolean otelActiveFromCustomResource = config.getAgentOtelActive();
+    Integer otelPortFromCustomResource = config.getAgentOtelPort();
+
+    if(otelActiveFromCustomResource || getBoolean(otelActiveFromEnvVar)){
+      Integer otelPort = otelPortFromCustomResource;
+      if(isInteger(otelPortFromEnvVar)){
+        otelPort = getInteger(otelPortFromEnvVar);
+      }
+      container.getPorts().add(new ContainerPort(otelPort,null, null,null,null));
```
mdonkers, line 363: On the Agent-side, the OTel port is fixed. So while we could expose this under a different port-number, the `containerPort` should always be the same.
I'll see if we can successfully change the `hostPort` though, so someone could expose it differently.

mdonkers (changes requested):
Thanks for these contributions, this is really helpful.

There are a few minor things to fine-tune which I have commented on. But I'll have a closer look myself as well.

Probably important to note is that we had to add a DCO bot for sign-off on the commits. Not sure if you see it already, otherwise probably soon (might be with a new commit already). In any case, I would like to kindly ask you to sign-off on all the commits (would require a rebase), so we can smoothly merge this. See here: https://github.com/probot/dco#how-it-works for any guidance, or let us know if you need further assistance.

src/main/java/com/instana/operator/AgentDeployer.java
asafarian: Ok, its changed.

src/main/java/com/instana/operator/customresource/InstanaAgentSpec.java
asafarian: Ok

src/main/java/com/instana/operator/AgentDeployer.java
asafarian: should i remove this feature for the meanwhile ?
I don't think that we really need it.

mdonkers: Yes, I believe that would be easier for now if you don't need it.

asafarian: Okay, i am going to remove it..
Probably I will need to create a new PR, since i need to get rid of the old commits.

asafarian:
@mdonkers, Hey Miel I just made the changes and squashed the commits for the 2 Support Tickets I created.
Please check and let me know, if there is anything missing. 
The code needs to be tested on a K8 Cluster.

src/main/java/com/instana/operator/AgentDeployer.java
asafarian: Ok, its done. I reduced it to 2 commits. One Commit per support Ticket.

src/test/resources/customresource-max.yaml
```diff
@@ -11,12 +11,15 @@ agent.daemonSetName: test-daemon-set
...
 agent.rbac.create: false
 agent.image: instana/test-image:1.2.3
+agent.imagePullPolicy: Always
 agent.cpuReq: 0.7
 agent.cpuLimit: 1.8
 agent.memReq: 513
 agent.memLimit: 518
 agent.downloadKey: test-download-key
 agent.host.repository: /Users/stan/.m2/repository
+agent.otel.active: true
```
mmanciop, line 21: imo, change simply to `opentelemetry.enabled: <boolean>`

src/test/resources/customresource-max.yaml
```diff
@@ -11,12 +11,15 @@ agent.daemonSetName: test-daemon-set
...
+agent.otel.port: 55680
```
mmanciop, line 22: I do not think it makes sense to make this configurable.

mdonkers:
@asafarian , You might have seen some feedback from @mmanciop . I'm handling this in a separate commit / PR (depending on what is possible). I'm currently verifying but also including some additional improvement. So bear with me just a bit longer please :-)

mdonkers (approved):
Merging, and doing some small updates in a separate PR.

PR #61, opened 2021-06-21
Build multi-arch images on Concourse CI

htreu:
This new [Concourse CI pipeline](https://ci.instana.io/teams/agent/pipelines/agent-operator-release) 
* builds the current Java based operator from release tags (`v1.0.7`) and pre-release tags (`v1.0.8-pre` or `v1.0.8-<your-name-here>`)
* builds multi-architecture docker images (amd64, arm64, s390x) on AWS CloudBuild
* publishes internal to GCR
* for release tags: publishes to docker.io (multi-arch) & Red Hat Container Registry (only amd64, [details here](https://access.redhat.com/solutions/5583611))
* builds the OLM bundle
* builds OLM docker images and publishes internal to GCR
* for release tags: publishes OLM to Red Hat Container Registry

### DIFF
--- a/ci/pipeline.yaml
+++ b/ci/pipeline.yaml
@@ -20,7 +20,7 @@ resources:
       branch: main
       #  match release tags like "v1.12.99"
       #  match pre-release tags like "v1.12.100-pre"
-      tag_regex: ^v\d*\.\d*\.\d*.*$
+      tag_regex: ^v\d+\.\d+\.\d+.*$
       #          ^ beginning of the tag string
       #           ^ all tags start with "v"
       #            ^ ^ any number of digits
--- a/ci/pipeline.yaml
+++ b/ci/pipeline.yaml
@@ -385,18 +385,20 @@ jobs:
                   exit 0
                 fi
 
-                # According to this knowledge base article https://access.redhat.com/solutions/5583611 the RH container registry does not support fat manifest lists.
-                # For now we will fall back to just publish the amd64 variant.
-
                 echo "---> re-tagging images for Red Hat Container Registry"
-                docker tag "gcr.io/instana-agent-qa/instana-agent-operator:$COMMIT_SHA-$VERSION-amd64" "$RED_HAT_REGISTRY:$OPERATOR_DOCKER_VERSION"
-                # docker tag "gcr.io/instana-agent-qa/instana-agent-operator:$COMMIT_SHA-$VERSION-s390x" "$RED_HAT_REGISTRY:$OPERATOR_DOCKER_VERSION-s390x"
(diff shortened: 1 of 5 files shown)


dlbock:
LGTM as long as it works for you 💪 My only point of feedback is to pull the task related shell code that's currently embedded directly into `ci/pipeline.yml` into task files  in `ci/tasks/...`, and maybe even pull the shell code into script files and reference it from the task files. Currently `ci/pipeline.yml` is 500+ lines long.

ci/pipeline.yaml
```diff
@@ -0,0 +1,578 @@
...
+      #  match pre-release tags like "v1.12.100-pre"
+      tag_regex: ^v\d*\.\d*\.\d*.*$
+      #          ^ beginning of the tag string
+      #           ^ all tags start with "v"
+      #            ^ ^ any number of digits
+      #               ^ the "." character
+      #                 ^ ^ any number of digits
+      #                    ^ the "." character
+      #                      ^ ^ any number of digits
+      #                         ^^ any prefix (denoting a non-release tag)
```
mdonkers, line 31: Love the description.
nitpick: shouldn't it be `\d+` everywhere? Because a `*` is zero to many, a `+` would ensure at least a single digit.

htreu:
@dlbock I thought about it. but then during development of the pipeline it would require a new commit for every change to the shell scripts. Right now I can do `fly ... set-pipeline ...` and try again.

htreu:
Unfortunately modularization is not the core competence of Concourse.

ci/pipeline.yaml
```diff
@@ -0,0 +1,578 @@
...
+                then
+                  echo "Manifest $RED_HAT_REGISTRY:${OPERATOR_DOCKER_VERSION} is already present on Red Hat Container Registry, skipping publish."
+                  exit 0
+                fi
+
+                # According to this knowledge base article https://access.redhat.com/solutions/5583611 the RH container registry does not support fat manifest lists.
+                # For now we will fall back to just publish the amd64 variant.
+
+                echo "---> re-tagging images for Red Hat Container Registry"
+                docker tag "gcr.io/instana-agent-qa/instana-agent-operator:$COMMIT_SHA-$VERSION-amd64" "$RED_HAT_REGISTRY:$OPERATOR_DOCKER_VERSION"
```
mdonkers, line 392: Question; Not sure here; for consistency didn't we tag the `amd64` version also with `-amd64` and publish it?

mdonkers (approved):
minor remarks, other than that LGTM

dlbock:
> Unfortunately modularization is not the core competence of Concourse.

I agree

ci/pipeline.yaml
htreu: Our agent images in fact do get published with all `-<arch>` suffixes. I will update here to also include the multi-arch variants, but instead of pushing the manifest list fall back to the amd64 image.

PR #62, opened 2021-06-23
rewrite leader election in Golang

ucf91:


controllers/leaderelection/agent_leaderelection.go
```diff
@@ -0,0 +1,181 @@
...
+	LeaderElectionTask, err = LeaderElectionTaskScheduler.ScheduleWithFixedDelay(func(ctx context.Context) {
+		var activePods map[string]coreV1.Pod
+		var err error
+		if activePods, err = l.fetchPods(agentNameSpace); err != nil {
+			l.Log.Error(err, "Unable to fetch agent pods for doing election")
+			return
+		}
+		l.pollAgentsAndAssignLeaders(activePods)
+
+		time.Sleep(5 * time.Second)
```
mdonkers, line 51: Why sleeping here 5 seconds? This gets scheduled so sleeping should not be needed and is overhead?

controllers/leaderelection/agent_leaderelection.go
```diff
@@ -0,0 +1,181 @@
...
+	}, 5*time.Second)
+	if err == nil {
+		l.Log.Info("Task has been scheduled successfully.")
```
mdonkers, line 54: Perhaps logging the error (if any) is even more important than logging when succeeded?

controllers/leaderelection/agent_leaderelection.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	return err
+}
+
+func (l *LeaderElector) CancelLeaderElection() {
+	if LeaderElectionTask != nil && !LeaderElectionTask.IsCancelled() {
+		LeaderElectionTask.Cancel()
+	}
+	LeaderElectionTaskScheduler.Shutdown()
```
mdonkers, line 63: Shouldn't also check if `LeaderElectionTaskScheduler` is `nil`? Might not have been initialized if `StartCoordination` was never called before

controllers/leaderelection/agent_leaderelection.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+	}
+	return activePods, nil
+}
+
+func (l *LeaderElector) pollAgentsAndAssignLeaders(pods map[string]coreV1.Pod) {
+	for {
+		leadershipStatus := l.pollLeadershipStatus(pods)
+
+		if len(leadershipStatus.Status) == 0 {
```
mdonkers, line 90: Question, what does this mean, what are you checking exactly? (I see its an empty list, but what does that mean? comment might help here)

controllers/leaderelection/agent_leaderelection.go
ucf91: I can add check for it as well, but it can't happen as the start coordination must be called first in reconcile

controllers/leaderelection/agent_leaderelection.go
ucf91: if leadershipStatus.Status is empty list that means we were not able to get leadershipstatus for any pod so we can't proceed with leadership assigning process and we should return, I added that now as a comment

PR #67, opened 2021-09-23
Add TLS encryption

tkohn:
Enabled TLS encryption in two variants: 

1. use an existing secret of type `kubernetes.io/tls` as a reference.
2. provide a certificate and a private key from which a secret of type `kubernetes.io/tls` will be created.


mdonkers (approved):
LGTM,

Only minor question regarding creating a Secret from crt and key. I noticed secrets are normally stored in K8s with base64 encoding. We'd expect these to be provided by the customer already encoded, right?

tkohn:
Yes that is correct. 🤔  alternative using the `|` would be a solution and do the encoding in the code.
For helm I mentioned it in the README and documentation, that the value has to be base64 encoded.

PR #69, opened 2021-09-29
Configure memory limit/request dependent on node architecture ppc

tkohn:
# Improve memory configuration for ppc-architecture

## Why

The agents consume more memory on a cluster with ppc-architecture

## What

Increase memory limit/request for agent running on ppc-architecture

### DIFF
--- a/src/main/java/com/instana/operator/AgentDeployer.java
+++ b/src/main/java/com/instana/operator/AgentDeployer.java
@@ -49,6 +49,9 @@ public class AgentDeployer {
 
   private static final String DAEMON_SET_NAME = "instana-agent";
   private static final String VERSION_LABEL = "app.kubernetes.io/version";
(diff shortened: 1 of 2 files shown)


src/main/java/com/instana/operator/AgentDeployer.java
```diff
@@ -300,14 +303,15 @@ DaemonSet newDaemonSet(InstanaAgent owner,
       container.setResources(new ResourceRequirements());
     }
 
+    final String arch = getArchFromNodeInfo(owner.getMetadata().getNamespace());
```
tkohn, line 306: I am not sure about the arch variable. Maybe it would be better to move it to customResourceAdded and set it once.
Or it would be perfect to get the architecture of node where the agent will be run.

htreu (approved):
lgtm

PR #74, opened 2021-12-03
Fix container issue with SecurityContext and add some Bundle improvements

mdonkers:


Dockerfile
```diff
@@ -68,6 +68,7 @@ COPY LICENSE /licenses/
 
 RUN mkdir -p .cache/helm/repository/
 RUN chown -R ${USER_UID}:${USER_UID} .cache
+RUN chmod -R 777 .cache
```
tkohn, line 71: `777` is it necessary to have read/write/execute for all

tkohn (approved):

Dockerfile
mdonkers: correct.
We removed the default `securityContext`. What happens on OpenShift is it will create its own `securityContext` with a random user id. The executed code then runs as this user, which is a _different_ user than the `1001` id that created the `.cache` folder and has the owner permissions. So then the running code, because of not having the `write` permissions, can't download and create the Helm charts to the cache.

PR #82, opened 2022-03-22
Update image path to use icr.io registry

tkohn:
- use go 1.18 builder image to build operator
- remove yum and replace it with microdnf update


Dockerfile
```diff
@@ -62,9 +62,7 @@ ENV OPERATOR=instana-agent-operator \
     USER_UID=1001 \
     USER_NAME=instana-agent-operator
 
-RUN microdnf install yum \
```
tkohn, line 65: That command install additional packages which will increase the number of found CVEs

PR #88, opened 2022-11-01
add arch and os affinities and CSV labels

zach-robinson:


FelixMarxIBM:
I noticed that I did not clean up both instana-agent-operator.clusterserviceversion.yaml files with my commit in
17918570a5a27fb28333293a8b43a84b95ae1798

Do you happen to know what the correct handling is here if we want to have them in sync?
The correct installModes are

```
  installModes:
  - supported: false
    type: OwnNamespace
  - supported: false
    type: SingleNamespace
  - supported: false
    type: MultiNamespace
  - supported: true
    type: AllNamespaces
```

Is it fine to just not mention those if we want false? If so your changes are fine and we can get rid of the wrong values.

zach-robinson:
@FelixMarxIBM I assume they default to false if omitted, but I suppose we might as well set them to false explicitly just to be sure.

zach-robinson:
Looks like when you run the `operator-sdk generate bundle` command, it will explicitly set the unlisted ones to false in the generated CSV (the complete CSV in `/bundle` which is auto generated based on the contents of `/config`), so we should be fine with the way it is now.

tkohn (approved):

PR #93, opened 2023-01-10
Auto prs

zach-robinson:


ci/rh-olm-release-pipeline.yaml
```diff
@@ -0,0 +1,400 @@
...
+                          microdnf install git zip curl jq
+                          
+                          pushd instana-agent-operator-release
+                          
+                          OLM_BUNDLE_ZIP=$(ls olm*.zip)
+                          OPERATOR_RELEASE_VERSION=$(echo $OLM_BUNDLE_ZIP | sed 's/olm-\(.*\)\.zip/\1/')
+                          COMMIT_MESSAGE="operator instana-agent-operator ($OPERATOR_RELEASE_VERSION)"
+                          OLM_BUNDLE_ZIP_PATH="$(pwd)/$OLM_BUNDLE_ZIP"
+                          
+                          # If the PR is already open then we are done
```
tkohn, line 99: Nice check 💪

ci/rh-olm-release-pipeline.yaml
```diff
@@ -0,0 +1,400 @@
...
+
+resources:
+  - name: agent-operator-pipeline-source
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/instana/instana-agent-operator.git
+      branch: main
+
+  - name: instana-agent-operator-release
```
tkohn, line 11: I have seen that a resource rate limit happens for this and for the `yq-release`.

One possibility would be to use a token to get a higher limit. Or we can set `check_every` and change the check to every hour or a higher value. The default is one minute, and there are certainly many resources outside our team.
So it is not this pipeline, it's the total number of resources using the Github API
https://concourse-ci.org/resources.html#schema.resource.check_every

ci/rh-olm-release-pipeline.yaml
```diff
@@ -0,0 +1,400 @@
...
+    type: github-release
+    source:
+      owner: instana
+      repository: instana-agent-operator
+
+  - name: yq-release
```
tkohn, line 17: see above

ci/rh-olm-release-pipeline.yaml
```diff
@@ -0,0 +1,400 @@
...
+                          fi
+                          
+                          popd
+                          
+                          pushd community-operators-repo/operators/instana-agent-operator
+                          
+                          mkdir -p $OPERATOR_RELEASE_VERSION
+                          unzip -o $OLM_BUNDLE_ZIP_PATH -d $OPERATOR_RELEASE_VERSION
+                          
+                          git config --global user.name "instanacd"
```
tkohn, line 114: we should not hardcoded username and email. Please use the Secret Manager. Also it is public github repository, so we have to hide these informations.

tkohn (changes requested):
- Please try to refactor the duplication into a separate script file. This would make it easier to change the logic for all at once.
- Also hide email and username of the used user

I like the improvements and the automation was long overdue. Really good work and thank you for it.

ci/rh-olm-release-pipeline.yaml
zach-robinson: Concourse requires us to bind secrets like this: `((instanacd-github-api-token))`, so the user is already hardcoded and visible in a sense, isn't it? I can make `((instanacd-github-username))` and `((instanacd-github-email))`, but that information will be visible anyway by looking at the instancd github account which will be attached to the PRs that get opened...

tkohn: You are right about both. One benefit for using Secret Manager would be rotation of both entries.
And with the naming, we could use a more generic name without the instanacd. I will ask tomorrow other teams about the naming, maybe it is okay to hardcode both

tkohn (approved):

PR #102, opened 2023-04-05
Bump go version

Milica-Cvrkota-IBM:


FelixMarxIBM:
Let's merge this after the rbac change, so it includes only the verison bump.

Milica-Cvrkota-IBM:
This was just for testing. Created card on the board to do investigation, we need more changes

PR #105, opened 2023-05-22
Adding resource limits example

IsReal8a:
Hi team,
Some customers and including myself needed an example on how to define resource limits (CPU/MEM) for the agent, I have added some definitions in the sample YAML file.

### DIFF
--- a/config/samples/instana_v1_instanaagent.yaml
+++ b/config/samples/instana_v1_instanaagent.yaml
@@ -16,3 +16,8 @@ spec:
     configuration_yaml: |
       # You can leave this empty, or use this to configure your instana agent.
       # See https://docs.instana.io/setup_and_manage/host_agent/on/kubernetes/
+  # Example on how to limit Agent Resources
+  agent.cpuLimit: "1.0"
+  agent.cpuReq: "0.5"
+  agent.memLimit: "512Mi"
+  agent.memReq: "256Mi"


Milica-Cvrkota-IBM:
I am not sure we want to put the custom values for the resource limits in the "default" configuration. Would it be useful if we put more explicit pointer to the list of parameters in the docs https://github.ibm.com/instana/instana-agent-charts/tree/main/canonical#configuration-reference , with the complete list of parameters?

config/samples/instana_v1_instanaagent.yaml
```diff
@@ -16,3 +16,8 @@ spec:
     configuration_yaml: |
       # You can leave this empty, or use this to configure your instana agent.
       # See https://docs.instana.io/setup_and_manage/host_agent/on/kubernetes/
+  # Example on how to limit Agent Resources
```
FelixMarxIBM, line 19: Are we sure these are correct and will be supported in the future.
If I look at the helm chart, they are referenced like this:
https://github.ibm.com/instana/instana-agent-charts/tree/main/canonical#configuration-reference:~:text=agent.pod.limits,512Mi


```
agent.pod.limits.cpu
agent.pod.limits.memory
agent.pod.requests.cpu
agent.pod.requests.memory
```

IsReal8a: Interesting, I found this information on the operator and I applied the changes in my the monitoring without errors.
https://github.com/instana/instana-agent-operator/blob/main/config/crd/bases/instana.io_agents.yaml

IsReal8a:
I do agree Milica as I didn't know the existence of the agent-charts... If you look to the [official documentation](https://www.ibm.com/docs/en/instana-observability/current?topic=agents-configuring-host#limiting-host-agent-cpu-and-memory), in Kubernetes is not clear where to apply that configuration and where to get more details about the changes that need to happen to make that work. I needed to dig in the code to realize that, not even Google pointed to the agent-charts.

Either:

- We create a complete sample file for easy-to-use purpose for the customer or 
- We point the official documentation to the agent-charts configuration for reference.

Please note that the configuration I used it and shared by this commit was accepted by the operator.

How do we proceed from here?

I can help in any way if needed. Thanks.

Milica-Cvrkota-IBM:
Hi @IsReal8a ,
In the official docs, we have sections where we provide the info how to do this:
https://www.ibm.com/docs/en/instana-observability/current?topic=agents-installing-host-agent-kubernetes#install-operator-manually & https://www.ibm.com/docs/en/instana-observability/current?topic=agents-installing-host-agent-kubernetes#operator-configuration, and the link to the helm configuration is provided as well.

We won't merge this PR because the example file here only contains the required params, and we want to keep it that way. We will create another sample file in addition with the extended list of the params. Closing this

PR #110, opened 2023-09-21
remove containers.instana.io login

Milica-Cvrkota-IBM:


konrad-ohms (approved):
LGTM, no need to login, if we don't use the registry in the build 👍

PR #119, opened 2023-11-22
ClusterRoles are not namespaced

diseku:


konrad-ohms:
Thanks for pointing out, after a bigger refactoring to a golang-based operator, the namespace was already removed, therefore closing the PR.

PR #137, opened 2024-04-11
fix: Adjust kubebuilder to fix cluster role

konrad-ohms:
The ClusterRole produced by the kubebuilder comments when running `make manifests` was out of sync with the definition in the CustomResourceDefinition. 

Whenever the operator was started locally via `make run`, the permissions from the local kubeconfig were inherited and worked fine, but starting from an image inside the cluster which relied on the serviceaccount mapped to the manger-role ClusterRole failed with forbidden.
This can be reproduced by running: `kubectl get
--as=system:serviceaccount:instana-agent:instana-agent-operator instanaagent.instana.io/instana-agent -n instana-agent`

Also worth noting: There is a custom verb `bind` for `rbac.authorization.k8s.io`. It must be set in this context, otherwise the operator is not allowed to bind a `clusterrole` to a `serviceaccount`, even if it has the create permission for the `clusterrolebinding` resource. This is necessary, as the agent operator will provision a ServiceAccount, a ClusterRole and a ClusterRoleBinding for the k8s_sensor.

### DIFF
--- a/controllers/instanaagent_controller.go
+++ b/controllers/instanaagent_controller.go
@@ -1,6 +1,6 @@
 /*
- * (c) Copyright IBM Corp. 2021
- * (c) Copyright Instana Inc. 2021
+ * (c) Copyright IBM Corp. 2021, 2024
+ * (c) Copyright Instana Inc. 2021, 2024
  */
 
 package controllers
@@ -95,7 +95,6 @@ func (r *InstanaAgentReconciler) reconcile(
 	log.Info("reconciling Agent CR")
 
 	agent.Default()
-
 	operatorUtils := operator_utils.NewOperatorUtils(ctx, r.client, agent)
 
 	if handleDeletionRes := r.handleDeletion(ctx, agent, operatorUtils); handleDeletionRes.suppliesReconcileResult() {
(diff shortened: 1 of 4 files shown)


controllers/instanaagent_controller.go
```diff
@@ -126,13 +125,25 @@ func (r *InstanaAgentReconciler) reconcile(
 	return reconcileSuccess(ctrl.Result{})
 }
 
-// +kubebuilder:rbac:groups=agents.instana.io,resources=instanaagent,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=instana.io,resources=agents,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=apps,resources=daemonsets;deployments,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=secrets;configmaps;services;serviceaccounts,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings,verbs=get;list;watch;create;update;patch;delete;bind
```
konrad-ohms, line 131: This change is important, otherwise the ClusterRole cannot be bound to the k8s serviceaccount

Milica-Cvrkota-IBM (approved):
LGTM 👍 
Leftover:
* check which SA agent uses, how does it work
* ensure the same SA is used for k8sensor and the agent

PR #140, opened 2024-04-25
Add minReadySeconds feature to agent daemonset and k8s-sensor deployment

FredrikAtIBM:


konrad-ohms (approved):
looks good, but we might want to check manually, if the properties are propagated to the daemonset and the deployment correctly in a real cluster.

konrad-ohms:
If we get https://github.com/instana/instana-agent-operator/pull/137 merged in before, we could even run it directly in a cluster and test it end-to-end.

PR #151, opened 2024-05-14
Add new required annotations

Milica-Cvrkota-IBM:
* currently public PRs for marketplaces cannot be merged because we are missing some required annotations: https://access.redhat.com/documentation/en-us/red_hat_software_certification/2024/html-single/red_hat_openshift_software_certification_policy_guide/index#con-operator-requirements_openshift-sw-cert-policy-products-managed
* move over Makefile logic from the `operator_3.0` branch


Makefile
```diff
@@ -71,12 +71,8 @@ help: ## Display this help.
 setup: ## Basic project setup, e.g. installing GitHook for checking license headers
 	cd .git/hooks && ln -fs ../../.githooks/* .
 
-
-##@ Development
-PYTHON3=python3
-manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+manifests: #controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
-	$(PYTHON3) ./hack/customize_crds.py
```
konrad-ohms, line 79: Not sure, if we require this post-processing still somehow, if not, I would also delete the file hack/customize_crds.py.
But it is gone in operator_3.0 anyway

konrad-ohms (approved):

konrad-ohms:
Unit tests and DCO are failing right now, controller-gen cannot be found

Makefile
Milica-Cvrkota-IBM: thanks, delete the file too

Milica-Cvrkota-IBM: It's not used anywhere else so should be safe to remove it

konrad-ohms (approved):
Looks good, the only fear I have is the number of merge conflicts we might see when we are going to merge operator_3.0

PR #162, opened 2024-05-28
instana_v1_extended_instanaagent: increase cpu requests and memory limits of k8sensor

chibby0ne:
Previously the cpu request were by default at 1% of a CPU. This meant that in every deployment the CPU usage was always more than the minimum CPU resource scheduled.

The memory limit which is the maximum before the process is killed with an OOM is increased to 2GB to be conservative with clusters with many updates and resources, where memory usage usually hovers around 1.3G


Milica-Cvrkota-IBM (approved):
LGTM 👍 We plan on releasing the operator tomorrow

PR #167, opened 2024-06-11
Use Fyre OCP clusters

Milica-Cvrkota-IBM:
* add another OCP cluster
* use Fyre clusters
* ensure that release of locks happens even if `cleanup-resources` fails


konrad-ohms (approved):

Milica-Cvrkota-IBM:
Aborting as we can't use Fyre in Concourse

Issue #170, opened 2024-06-19
The "controller-manager" Pods get OOMKilled in 2.1.0


bitgully:
We have been running the Instana Agent Operator version 2.0.32 successfully.
After upgrading to 2.1.0, the controller-manager pods run out of memory.
In the older version, the memory usage was around 180Mi (below the Deployment's default limit of 200Mi).
In 2.1.0, the memory usage we observed was up to 350Mi and hence above the limit.

- Is there a memory-related bug in 2.1.0?
- Can the Deployment's limit be increased to 400Mi?
- Should an option for custom memory limits be provided?

bitgully:
Fix has been released in version 2.1.1. Thanks!

konrad-ohms:
Happy to help, thanks for reporting :slightly_smiling_face:

PR #171, opened 2024-06-20
chore: Increase memory limit of controller manager

konrad-ohms:
Allow up to 600Mi memory consumption, even if only 200Mi get requested upfront. Only the leader pod might exceed the 200Mi and the real amount of memory might depend on the environment, therefore adding some buffer.

See https://github.com/instana/instana-agent-operator/issues/170 and TS016461105

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -64,7 +64,7 @@ spec:
         resources:
           limits:
             cpu: 200m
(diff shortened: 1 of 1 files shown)


Milica-Cvrkota-IBM (approved):

konrad-ohms:
Changing target branch to combine multiple changes into a single release

PR #175, opened 2024-06-27
doc(ci): Describe manual operator catalog promotion

konrad-ohms:
In case the CI promotion fails, document how to create PRs manually.

Should be safe to merge, I placed the markdown in the ci folder which will not trigger a new release automatically.


FelixMarxIBM (approved):
LGTM

PR #180, opened 2024-07-02
Feat: Additional-backend support for host agent

konrad-ohms:
In some setups, the agent supports to communicate with multiple instana backends.
The first step is to enable the agent host sensor to send data to multiple backends which is already supported in non-operator based deployments.
This change will allow the configuration of additional backends.
As backend configuration will contain secret information, those should not be stored in configmaps, but in Kubernetes secrets.

Requires https://github.ibm.com/instana/instana-agent-docker/pull/16 to be merged first.
Work item: https://jsw.ibm.com/browse/INSTA-8755

Example CR:

```yaml
apiVersion: instana.io/v1
kind: InstanaAgent
metadata:
  name: instana-agent
  namespace: instana-agent
spec:
  zone:
    name: konrad-gke
  cluster:
    name: konrad-operator-e2e-test-dev-cluster
  agent:
    keysSecret: instana-agent-key
    endpointHost: ingress-red-saas.instana.io
    endpointPort: "443"
    additionalBackends:
    - endpointHost: ingress-magenta-saas.instana.rocks
      endpointPort: "443"
      key: xxx
    env: {}
    configuration_yaml: |
      # See https://docs.instana.io/setup_and_manage/host_agent/on/kubernetes/

```


konrad-ohms:
Find rebased version in https://github.com/instana/instana-agent-operator/pull/200

PR #185, opened 2024-07-09
add missing inputs

Milica-Cvrkota-IBM:


konrad-ohms (approved):
Looks good, but please check the DCO

ci/scripts/rebase-fork-branches.sh
```diff
@@ -25,19 +25,19 @@ rebase_branches() {
     echo "Rebasing the forked main and the found PR branches from the upstream main branch"
     pushd "$PUBLIC_REPO_LOCAL_NAME"
     echo "Rebasing the forked main"
-    git remote add upstream https://github.com/"$OWNER"/"$REPO".git
+    git config --global user.name "instanacd"
+    git config --global user.email "instanacd@instana.com"
+    git remote add upstream https://github.com/"$OWNER"/"$REPO".git 2>/dev/null || echo "Remote upstream already exists"
```
konrad-ohms, line 30: Not entirely sure, but as you have `set -e` at the top, would adding an existing remote not terminate the entire script, as it is not existing with 0?

ci/scripts/commit-changes-to-public-repo.sh
```diff
@@ -37,7 +37,7 @@ popd
 pushd "$PUBLIC_REPO_LOCAL_NAME"/operators/"$OPERATOR_NAME"
 
 echo "Creating a new feature branch from the upstream main"
-git remote add upstream https://github.com/"$OWNER"/"$REPO".git
+git remote add upstream https://github.com/"$OWNER"/"$REPO".git 2>/dev/null || echo "Remote upstream already exists"
```
konrad-ohms, line 40: Not entirely sure, but as you have set -e at the top, would adding an existing remote not terminate the entire script, as it is not existing with 0?

ci/scripts/rebase-fork-branches.sh
Milica-Cvrkota-IBM: Thanks for the hint, just tested this locally with a dummy script but it seems that the second part of OR gets executed successfully

PR #199, opened 2024-07-17
Move creation of PR to the commit-changes task

Milica-Cvrkota-IBM:
With the new logic we don't need to create PRs in a separate task as there is currently a lot of repetition. As we need to pass the feature branch to `curl -fX POST` command, we can just perform the task within the same `commit-changes` task


FelixMarxIBM (approved):
This looks much cleaner

PR #203, opened 2024-07-23
Extend list of files that should be ignored for triggering release

Milica-Cvrkota-IBM:


ci/scripts/tag-new-release.sh
```diff
@@ -25,9 +25,9 @@ fi
 
 only_ci_changes=true
```
FelixMarxIBM, line 26: Maybe use individual ifs for better readability?

```
only_ci_changes=true
for file in $(git diff --name-only "$latest_release"..HEAD); do 
    if [[ $file == ci/* ]]; then
        continue
    fi
    
    if [[ $file == .* ]]; then
        continue
    fi
    
    if [[ $file == *.md ]]; then
        continue
    fi
    
    if [[ $file == Makefile ]]; then
        continue
    fi

    echo "Found file that is not in ci/ directory, it's not a hidden file/directory, an .md file nor Makefile: $file"
    only_ci_changes=false
    break
done
```

Milica-Cvrkota-IBM: That sounds good, updated the PR

FelixMarxIBM (approved):

PR #204, opened 2024-08-01
Bugfix: Fix issue where the CRD Status was reporting false-positive

FredrikAtIBM:


konrad-ohms (approved):
good catch, but looks like the branch name is causing issues when the e2e image is pushed with the branch name as tag.
Please avoid slashes in branch names, or we could consider adjusting the pipeline logic.

PR #205, opened 2024-08-01
CI: Add e2e test case to check for CR state Running

konrad-ohms:
Addition to change: https://github.com/instana/instana-agent-operator/pull/204


FelixMarxIBM (approved):
LGTM

PR #206, opened 2024-08-05
WIP: Rewrite e2e tests to golang

konrad-ohms:
Right now we have a shell script executing e2e tests.
I would consider re-implementing them in pure go to keep them extensible in the long run.

While searching for existing solutions, I found that there is an e2e framework available at the Kubernetes SIGs: https://github.com/kubernetes-sigs/e2e-framework/tree/main/examples/real_cluster

We could either rely on that (which forced a few updates on dependencies) or we could fallback to pure command execs to mimic the bash behavior with a bit additional golang logic, e.g. for waiting on resources.


konrad-ohms:
rewritten as part of https://github.com/instana/instana-agent-operator/pull/217

PR #207, opened 2024-08-05
Feat: Multi-Backend support for host agent

konrad-ohms:
Adding support for the host agent to send data to multiple backends and add end-to-end tests for it.


FredrikAtIBM (approved):
LGTM

PR #209, opened 2024-08-06
Fix: Bind only enabled ports on agent pods

konrad-ohms:
Card: [INSTA-12524](https://jsw.ibm.com/browse/INSTA-12524)

- Remove obsolete agent-socket port 42666
- Do only bind ports which are enabled on the Daemonset

```
spec:
  agent:
    configuration_yaml: |
      # See https://github.com/instana/instana-agent-operator/blob/main/config/samples/instana_v1_extended_instanaagent.yaml for the extended version.
    endpointHost: xxx
    endpointPort: "443"
    env: {}
    key: xxx
  cluster:
    name: konrad
  opentelemetry:
    grpc:
      enabled: false
    http:
      enabled: true
```

If grpc is enabled, bind 42699 (agent apis) and 55680 (oltp-legacy) and 4317 (otlp-grpc) ports. 

If http is enabled, bind 42699 (agent apis) and 4318 (oltp-http) ports.

By default, bind all ports.

### DIFF
--- a/pkg/k8s/object/builders/common/ports/instana_agent_port.go
+++ b/pkg/k8s/object/builders/common/ports/instana_agent_port.go
@@ -28,8 +28,6 @@ type InstanaAgentPort string
 const (
 	AgentAPIsPort                 InstanaAgentPort = "agent-apis"
 	AgentAPIsPortNumber           int32            = 42699
-	AgentSocketPort               InstanaAgentPort = "agent-socket"
-	AgentSocketPortNumber         int32            = 42666
 	OpenTelemetryLegacyPort       InstanaAgentPort = "otlp-legacy"
 	OpenTelemetryLegacyPortNumber int32            = 55680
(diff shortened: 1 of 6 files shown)


pkg/k8s/object/builders/common/ports/instana_agent_port.go
```diff
@@ -60,18 +56,12 @@ func (p InstanaAgentPort) PortNumber() int32 {
...
-	case OpenTelemetryHTTPPort:
+	}
+	if p == OpenTelemetryHTTPPort {
 		return openTelemetrySettings.HttpIsEnabled()
-	case AgentAPIsPort:
-		fallthrough
-	case AgentSocketPort:
-		fallthrough
-	default:
-		return true
```
konrad-ohms, line 75: Not a behavior change, but I found it less easy to read with switch case, so changed it to multiple if clauses.

konrad-ohms:
replaced by https://github.com/instana/instana-agent-operator/pull/211

PR #210, opened 2024-08-08
ci: Rely on commit instead of branch for pulling images

konrad-ohms:
Caching might cause old images to being pulled if branch names are used


FredrikAtIBM (approved):
Looks like this addition will hit several birds with one stone

PR #214, opened 2024-08-19
chore: Add opentelemetry default settings to example

konrad-ohms:
When not defined, there is no default value shown on the OCP install page.


Milica-Cvrkota-IBM (approved):
LGTM

PR #218, opened 2024-09-09
go.mod: change version from 1.23 to 1.23.0

chibby0ne:
1.23 is not a downloable release. Go mod should list the specific toolchain that can be downloaded as mentioned here: https://github.com/golang/go/issues/62278#issuecomment-1693538776 Otherwise opening the project having a different version installed will result in errors, e.g:

```
$ go version
go: downloading go1.23 (linux/amd64)
go: download go1.23 for linux/amd64: toolchain not available
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/instana/instana-agent-operator
 
-go 1.23
+go 1.23.0
 
 require (
 	github.com/Masterminds/goutils v1.1.1


konrad-ohms (approved):
Thanks @chibby0ne, good catch

konrad-ohms:
I am still wondering why they document it differently: https://go.dev/doc/modules/gomod-ref
image
image

chibby0ne:
Mmm that's a very good catch!
Maybe that snippet was done whilst using a development release and so, they had to specify the toolchain using `GO_TOOLCHAIN`? 
Or maybe that snippet was done before this "go toolchain" thing was defined and so the `go` cli wouldn't try to download the exact version written in the `go.mod` and so it wouldn't result in an error?
In any case, a case can be made of proposing a doc change with another more up to date code snippet with less surprising behavior if one were to recreate it.

PR #224, opened 2024-10-29
[INSTA-17884] feat: Increase default k8sensor pollrate to 10s

Ferenc-:
The previous value was 1 second.
And the intention with this change is to optimizing data volume, and also match the Kubernetes UIs granularity, that doesn't go below 10 seconds. Read more in INSTA-17881.

### DIFF
--- a/pkg/k8s/object/builders/k8s-sensor/deployment/deployment.go
+++ b/pkg/k8s/object/builders/k8s-sensor/deployment/deployment.go
@@ -140,7 +140,7 @@ func (d *deploymentBuilder) build() *appsv1.Deployment {
 							Image:           d.Spec.K8sSensor.ImageSpec.Image(),
 							ImagePullPolicy: d.Spec.K8sSensor.ImageSpec.PullPolicy,
 							Command:         []string{"/ko-app/k8sensor"},
-							Args:            []string{"10s"},
+							Args:            []string{"-pollrate", "10s"},
 							Env:             d.getEnvVars(),
 							VolumeMounts:    mounts,
 							Resources:       d.Spec.K8sSensor.DeploymentSpec.Pod.ResourceRequirements.GetOrDefault(),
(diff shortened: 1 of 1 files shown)


konrad-ohms:
@Ferenc- Thanks for your contribution, could you please add a short description to the PR why the default was changed or what was the default before?
As this is a public repository, customers might notice the change and might want to understand reasons for the change.

pkg/k8s/object/builders/k8s-sensor/deployment/deployment.go
```diff
@@ -139,6 +139,8 @@ func (d *deploymentBuilder) build() *appsv1.Deployment {
 							Name:            "instana-agent",
 							Image:           d.Spec.K8sSensor.ImageSpec.Image(),
 							ImagePullPolicy: d.Spec.K8sSensor.ImageSpec.PullPolicy,
+							Command:         []string{"/ko-app/k8sensor"},
+							Args:            []string{"10s"},
```
chibby0ne, line 143: I think this needs the `-pollrate` before the `10s`

Ferenc-: Very true. Done.

konrad-ohms:
@Ferenc-  is this branch still necessary?
If yes, please rebase it from main, as we changed the resource locking mechanism and your pipeline would not start to build otherwise.

Thanks.

Ferenc-:
> @Ferenc- is this branch still necessary?

Yes. It is now needed for INSTA-17884

 > If yes, please rebase it from main, as we changed the resource locking mechanism and your pipeline would not start to build otherwise.

Yes rebased.

Ferenc-:
The previous value was 1 second.
And the intention with this change is to optimizing data volume, and also match the Kubernetes UIs granularity, that doesn't go below 10 seconds. Read more in  INSTA-17881.

konrad-ohms:
Thanks for clarifying

Milica-Cvrkota-IBM (approved):

PR #226, opened 2024-11-14
Add roles for node metrics and stats for k8sensor

mansoorpervaiz:
K8sensor deployment needs updated ClusterRoles: `node/metrics` and  `nodes/stats`


konrad-ohms (approved):
As discussed, the k8sensor:latest image will gracefully handle situations where those permissions are not set. In deployments using the new operator, the new features will be leveraged.

PR #229, opened 2024-12-03
feat: support additional volume and mounts

boonware:
* Implements https://automation-management.ideas.ibm.com/ideas/INSTANA-I-935
* Allow users to define additional `volumes` and `volumeMounts` in the CRD. These can be useful for mounting files needed by specific sensors, for example, a keystore containing a TLS certificate needed to connect to a sensor target.


konrad-ohms (changes requested):
Thanks for your contribution, could you please add an example usage to https://github.com/instana/instana-agent-operator/blob/main/config/samples/instana_v1_extended_instanaagent.yaml ?
We try to keep examples in the repo for easy exploration of options on the CR.

e2e/extra_volume_test.go
```diff
@@ -0,0 +1,55 @@
...
+			err = decoder.ApplyWithManifestDir(ctx, r, "../config/samples", "instana_v1_extended_instanaagent.yaml", []resources.CreateOption{})
+			if err != nil {
+				t.Fatal(err)
+			}
+			t.Logf("CR created")
+
+			return ctx
+		}).
+		Assess("wait for first k8sensor deployment to become ready", WaitForDeploymentToBecomeReady(K8sensorDeploymentName)).
+		Assess("wait for second k8sensor deployment to become ready", WaitForDeploymentToBecomeReady(fmt.Sprintf("%s-1", K8sensorDeploymentName))).
```
konrad-ohms, line 48: In your test, there is no second backend configured. Therefore, this asses statement will fail and can be removed
```suggestion

```

config/samples/instana_v1_extended_instanaagent.yaml
```diff
@@ -101,6 +101,17 @@ spec:
...
+      # agent.pod.volumes and agent.pod.volumeMounts are additional volumes and volumeMounts for user-specific files.
+      # For example, a certificate may need to be mounted for an agent sensor to connect to the monitored target.
+      # https://kubernetes.io/docs/concepts/storage/volumes/
+      volumes:
+        - name: my-secret-volume
+          secret:
+            secretName: instana-agent-key
+      volumeMounts:
+        - name: my-secret-volume
+          mountPath: /secrets
```
konrad-ohms, line 114: When having a final look, I noticed that this is actually the extended_instanaagent.yaml which shows all available options, but typically just defines the defaults.
Your changes are adding new volumes which would not reflect the default behavior, so probably it would be more consistent to have your values in that file as comment and use a separate file for the test input.
But I also realized that your e2e test is relying on the given yaml to mount the secret, so I am fine for leaving it in as is for now, I might move the test input into a new file eventually, but that should not block the release for now.

konrad-ohms (approved):
Thanks, look good

PR #230, opened 2024-12-11
feat: support multiple configurations via configmaps

boonware:
## Description
Currently the operator supports only one definition of configuration YAML in the `InstanaAgent` custom resource. This PR adds support for defining multiple configurations and merging them.

## Problem
Consider the the problem of deploying the operator to monitor multiple DB2 instances in a cluster. To configure the DB2 plugin, the following configuration is needed:

```
com.instana.plugin.db2:
  remote:
    - ...
 ```
Note that this supports a list of configurations, one for each DB2 instance. This configuration may not be known at deployment time for the operator. Automation for deploying each DB2 instance must update this configuration by inserting its own details into the list, however, this will result in consistency problems if multiple automation instances try to update their details at the same time.

## Solution
* The problem described above is solved by allowing configuration sections to be defined in a `ConfigMap` with the special label `instana.io/agent-config=true`. The operator uses the Kubernetes API to find all such resources in the cluster, and merges their configuration into the `configuration_yaml` field read from `InstanaAgent` custom resource.
* The merge strategy for the `configuration_yaml` data is as follows:
    1. If the field is present in a config map but not present in the custom resource, then the field is copied from the config map into the custom resource.
    2. If the field is present in a config map and present in the custom resource, and if the type of the field is a list then the list entries from the config map are appended to the corresponding list in the custom resource.
    3. If the field is present in a config map and present in the custom resource, and if the type of the field is **not** a list, no change is made to the custom resource.
    4. If the field is present in the custom resource but not present in the config map, no change is made to the custom resource.
 
## Example
* Consider the `InstanaAgent` custom resource with the following YAML in the `configuration_yaml` field:
```
key1: value1
```
* Consider one `ConfigMap` with the following YAML in its `configuration_yaml` field under `data`:
```
key2: value2
com.instana.plugin.db2:
  remote:
    - host: host1
```
* Consider another `ConfigMap` with the following YAML in its `configuration_yaml` field under `data`:
```
com.instana.plugin.db2:
  remote:
    - host: host2
```
* The configuration YAML, as loaded by the operator, will be as follows:
```
key1: value1
key2: value2
com.instana.plugin.db2:
  remote:
    - host: host1
    - host: host2
```

### DIFF
--- a/controllers/apply.go
+++ b/controllers/apply.go
@@ -38,6 +38,7 @@ import (
 	"github.com/instana/instana-agent-operator/pkg/k8s/operator/operator_utils"
 	"github.com/instana/instana-agent-operator/pkg/k8s/operator/status"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
 func getDaemonSetBuilders(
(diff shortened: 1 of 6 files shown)


konrad-ohms:
@boonware is this branch still used?
If yes, please rebase it from main, as we changed the resource locking mechanism and your pipeline would not start to build otherwise.

Thanks.

PR #234, opened 2025-01-10
ci: Cleanup manual unlock

konrad-ohms:
Remove unneeded jobs from the main pipeline


Milica-Cvrkota-IBM (approved):

konrad-ohms:
only a pipeline change to main, this would not be tested as part of the test run, therefore merging without finished status checks.

PR #237, opened 2025-01-14
fix: Do not panic if number of volumes and mounts do not match

konrad-ohms:
Mounting the same secret multiple times is valid in Kubernetes, we should not panic here. Example:

```
  spec:
    agent:
      pod:
        volumeMounts:
        - mountPath: /opt/instana/agent/etc/application.jks
          name: jks-mount
          subPath: application.jks
        - mountPath: /custom-secrets
          name: jks-mount
        volumes:
        - name: jks-mount
          secret:
            secretName: keystore-secret-name
```


konrad-ohms:
@boonware FYI

Milica-Cvrkota-IBM (approved):

PR #243, opened 2025-02-04
Add ClusterRole/ClusterRoleBinding to instana-agent service account to enable prometheus sensor in Kubernetes

chibby0ne:
These two objects were left missing in the migration of the helm-chart from v1 to v2 and is affecting the correct operation of the prometheus sensor in Kuberentes environments.

In addition, there were a few rbac permissions missing in the `instana-agent-operator` service account, which are needed in the `instana-agent`, and since the `instana-agent-operator` cannot grant permissions to a service account that it tiself does not have, we are adding it there.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -6,6 +6,9 @@ metadata:
 rules:
 - nonResourceURLs:
   - /healthz
+  - /metrics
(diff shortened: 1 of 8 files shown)


Milica-Cvrkota-IBM (approved):
LGTM 👍

PR #245, opened 2025-02-04
Always pick up the latest changes for running e2e tests

Milica-Cvrkota-IBM:
* we use resource pipeline-source for running the e2e tests, but we only fetch new code if the `ci/` changes which is problematic
* the code change ensures we always fetch the latest changes of the repository, not just changes coming from the `ci/`


Milica-Cvrkota-IBM:
Merging without waiting for tests to pass as we are making the changes to the main release pipeline and this is not reflected in the PR tests

PR #251, opened 2025-02-28
fix: Handle missing proxy credentials correctly for k8sensor

konrad-ohms:
## Why

In case no credentials were defined, the HTTP proxy settings were rendered broken for the k8sensor. This change fixes this behavior and adds more unit tests.

## What

Add an additional check if credentials are present, remove them if not instead of render empty strings and keep the auth format.
Additionally, other the protocol setting should be translated into the env variable

## References

- INSTA-28191
- TS018624533

## Checklist

- [x] Backwards compatible?
- [ ] [Release notes](https://github.ibm.com/instana/docs/pull/13181/files) in public docs updated?
- [x] unit test coverage added or updated?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -27,8 +27,8 @@ For any larger changes this should include design choices.
 <!-- Please tick of these checklist items if applicable (or remove if not applicable). -->
 
 - [ ] Backwards compatible?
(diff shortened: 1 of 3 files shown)


FelixMarxIBM (approved):
LGTM

PR #254, opened 2025-04-10
Ensure list of env vars is same on reconcile

Milica-Cvrkota-IBM:
## Why

The operator was recreating pods when env vars were defined because the env var order changed on each reconcile, making the spec keep restarting until once they are aligned.

## What

We now sort the env vars by name before creating the pod spec which ensures the ordering is always consistent.

## References

- [Story / Card](https://jsw.ibm.com/browse/INSTA-33657)

## Checklist

- [x] Backwards compatible?
- [x] [Release notes](https://github.ibm.com/instana/docs/blob/main/src/pages/releases/agent_operator_notes/index.md) in public docs updated?


konrad-ohms (approved):
Good catch, thanks

konrad-ohms (approved):

PR #264, opened 2025-04-25
fix: Force k8sensor deployment updates on backend changes

konrad-ohms:
## Why

If there is a change in the additional backends, the k8sensor deployment currently can end up in situations in which the backend config is incorrect.

## What

Attach an annotation on the deployment pods with a checksum of the current backend. Whenever the backend config changes, the annotation gets adjusted which leads to a redeployment of affected pods.

## References

- [INSTA-31428](https://jsw.ibm.com/browse/INSTA-31428)

## Checklist

- [x] Backwards compatible?


pkg/k8s/object/builders/k8s-sensor/deployment/deployment.go
```diff
@@ -107,6 +111,28 @@ func addAppLabel(labels map[string]string) map[string]string {
...
+		endpointKeyFromSecret := d.keysSecret.Data["key"+d.backend.ResourceSuffix]
+		h.Write([]byte(d.backend.EndpointHost + d.backend.EndpointPort))
+		h.Write(endpointKeyFromSecret)
+	} else {
+		// backend secret was part of the CR
+		h.Write([]byte(d.backend.EndpointHost + d.backend.EndpointPort + d.backend.EndpointKey))
+	}
+
+	annotations["checksum/backend"] = fmt.Sprintf("%x", h.Sum(nil))
+	return annotations
```
konrad-ohms, line 133: This is the important change: Whenever there is a change in the CR, it ensures that a checksum for the backend is calculated and stored as annotation on the pod template.
This leads to a redeployment of existing pods to ensure the k8sensor gets refreshed environment variables.

pkg/k8s/object/builders/k8s-sensor/deployment/deployment.go
```diff
@@ -107,6 +111,28 @@ func addAppLabel(labels map[string]string) map[string]string {
...
+	// Deep copy annotations to extend them with a checksum
+	annotations := make(map[string]string, len(d.Spec.Agent.Pod.Annotations)+1)
+	for k, v := range d.Spec.Agent.Pod.Annotations {
+		annotations[k] = v
+	}
+
+	h := sha256.New()
+	if d.Spec.Agent.KeysSecret != "" {
+		// keysSecret contains the relevant data, if no key is found ignore it for the checksum
+		endpointKeyFromSecret := d.keysSecret.Data["key"+d.backend.ResourceSuffix]
```
Milica-Cvrkota-IBM, line 124: We could verify that d.keysSecret is set properly with nil check before access it to avoid potential panics

Milica-Cvrkota-IBM (approved):
Left one suggestion, otherwise LGTM 👍

pkg/k8s/object/builders/k8s-sensor/deployment/deployment.go
konrad-ohms: sounds good, thanks

PR #274, opened 2025-06-24
INSTA-13149: Make open telemetry ports and enablement more configurable

FredrikAtIBM:
## Why

Agent OTEL GRPC and HTTP ports should be configurable in a kubernetes environment using the operator.

## What

When the operator is configured to have open telemetry in its yaml:
image

The operator will prepare the things and pass over the port information to agent:
image

The agent will read the configurations appropriately and update to the new configuration:

GRPC
image

HTTP
image

- Legacy settings are respected. 
- By default, OTEL GRPC & HTTP port configurations are ENABLED with default ports 4317, 4318 respectively. 
- Changing a value explicitly will take precedence.
- Disabling the OTEL functionality through disabling the parent level OTEL field will result in everything becoming disabled.

## References

- [Story / Card](https://jsw.ibm.com/browse/INSTA-13149)

## Checklist

- [x] Backwards compatible?
- [ ] [Release notes](https://github.ibm.com/instana/docs/pull/13181/files) in public docs updated?
- [x] e2e test coverage added or updated?


api/v1/inline_types.go
```diff
@@ -382,44 +382,26 @@ type KubernetesDeploymentSpec struct {
...
 
+	// Specify whether GRPC is enabled (default is true).
 	// +kubebuilder:validation:Optional
-	GRPC *Enabled `json:"grpc,omitempty" yaml:"grpc,omitempty"`
+	GRPC OpenTelemetryPortConfig `json:"grpc,omitempty" yaml:"grpc,omitempty"`
 
+	// Specify whether HTTP is enabled (default is true).
 	// +kubebuilder:validation:Optional
-	HTTP *Enabled `json:"http,omitempty" yaml:"http,omitempty"`
+	HTTP OpenTelemetryPortConfig `json:"http,omitempty" yaml:"http,omitempty"`
```
konrad-ohms, line 395: Probably okay, but just want to confirm.
Is this a breaking change in that sense, that a user who defines grpc -> enabled=true will be able to upgrade the CRD?
Theoretically, it should work as the external representation did only add a port to an existing config, but at least internally it is a new struct now.
Did you try that?

api/v1/inline_types_test.go
```diff
@@ -80,232 +67,6 @@ func TestImageSpec_Image(t *testing.T) {
...
-			expected:              true,
-		},
-	} {
-		t.Run(test.name, func(t *testing.T) {
-			assertions := require.New(t)
-			assertions.Equal(test.expected, test.openTelemetrySettings.IsEnabled(), test.name)
-		})
-	}
-}
-
```
konrad-ohms, line 308: Should we extend the test suite with logic to test if ports are passed correctly instead of deleting it?

pkg/k8s/object/builders/common/ports/ports_builder_test.go
```diff
@@ -0,0 +1,209 @@
...
+			},
+			expectedServicePorts: []corev1.ServicePort{
+				ports.InstanaAgentAPIPortConfig.AsServicePort(),
+				ports.DefaultOpenTelemetryHTTPPortConfig.AsServicePort(),
+			},
+		},
+		{
+			name: "Disable all OLTP ports without legacy setting",
+			openTelemetrySettings: instanav1.OpenTelemetry{
+				Enabled: instanav1.Enabled{Enabled: pointer.To(false)},
```
konrad-ohms, line 82: Isn't line 82 the legacy config? The test name states it wanted to test to leave out the legacy config.

api/v1/inline_types_test.go
konrad-ohms: ah, saw it moved to ports_builder_test.go, never mind.

e2e/agent_test_api.go
```diff
@@ -704,8 +703,9 @@ func NewAgentCr() v1.InstanaAgent {
 				EndpointPort: strconv.Itoa(InstanaTestCfg.InstanaBackend.EndpointPort),
 			},
 			OpenTelemetry: v1.OpenTelemetry{
-				GRPC: &v1.Enabled{Enabled: &boolTrue},
-				HTTP: &v1.Enabled{Enabled: &boolTrue},
+				Enabled: v1.Enabled{Enabled: &enabled},
```
konrad-ohms, line 706: Do we need to define the legacy setting now by default? I thought we wanted to get rid of it, wherever possible.

konrad-ohms (changes requested):
Could we add an e2e test to actually see that the configuration could be applied?
Example:
- Deploy with changed ports
- When the deamonset is ready, `kubectl exec` into it and check that these are not opened: 

```
[root@worker1 agent]# ss -tulnp | grep 4318
tcp   LISTEN 0      4096   [::ffff:10.22.12.52]:4318             *:*    users:(("java",pid=299781,fd=218))                       
tcp   LISTEN 0      4096     [::ffff:127.0.0.1]:4318             *:*    users:(("java",pid=299781,fd=217))                       
[root@worker1 agent]# ss -tulnp | grep 4317
tcp   LISTEN 0      4096   [::ffff:10.22.12.52]:4317             *:*    users:(("java",pid=299781,fd=196))                       
tcp   LISTEN 0      4096     [::ffff:127.0.0.1]:4317             *:*    users:(("java",pid=299781,fd=194))
```

Instead we should be able to grep for your changed ports and see them being bound.

pkg/k8s/object/builders/agent/daemonset/daemonset.go
```diff
@@ -254,7 +260,7 @@ func (d *daemonSetBuilder) build() *appsv1.DaemonSet {
 									HTTPGet: &corev1.HTTPGetAction{
 										Host: "127.0.0.1",
 										Path: "/status",
-										Port: intstr.FromString(string(ports.AgentAPIsPort)),
+										Port: intstr.FromInt32(ports.InstanaAgentAPIPortConfig.Port),
```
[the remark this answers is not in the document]
konrad-ohms: It's actually not an issue, because this is a constant value: https://github.com/instana/instana-agent-operator/blob/d0e41ee6f64994d3257c52164391e6526f374fe6/pkg/k8s/object/builders/common/ports/instana_agent_port_config.go#L14

The struct defines it as type int32, so the struct can't be initialized as a string: https://github.com/instana/instana-agent-operator/blob/d0e41ee6f64994d3257c52164391e6526f374fe6/pkg/k8s/object/builders/common/ports/instana_agent_port_config.go#L35


The InstanaAgentAPIPortConfig is not changed anywhere in the code, they are defined as

```
var InstanaAgentAPIPortConfig = InstanaAgentPortConfig{
	Name: "agent-apis",
	Port: 42699,
}
```

The problem is that Go doesn't allow struct literals as constants because:
- Constants must be compile-time values
- Struct literals are considered composite types that are created at runtime
- Only basic types (strings, numbers, booleans) can be constants

We could consider making this more clear by doing something like:

```go
const (
    InstanaAgentAPIPortName = "agent-apis"
    InstanaAgentAPIPort     = 42699
)

var config = InstanaAgentPortConfig{
    Name: InstanaAgentAPIPortName,
    Port: InstanaAgentAPIPort,
}
```
or us a function to return static values
```go
func InstanaAgentAPIPortConfig() InstanaAgentPortConfig {
    return InstanaAgentPortConfig{
        Name: "agent-apis",
        Port: 42699,
    }
}
```

But to my knowledge there is no easy direct way to declare a struct as a constant.

konrad-ohms:
> A bit weird, but when working on the e2e tests it looks like that the changes keep the daemonset restarting pods, even on the commit [ed7df01](https://github.com/instana/instana-agent-operator/commit/ed7df0122b1538eae130fb1628723d5ea1756318) which was considered working by the pipeline. I need to followup on that problem before merging. Looks like the operator keeps reconciling for some reason.
> 
> The daemonset events look like this:
> 
> ```
> $ kubectl describe ds instana-agent
> ...
> Events:
>   Type    Reason            Age                From                  Message
>   ----    ------            ----               ----                  -------
>   Normal  SuccessfulCreate  82s                daemonset-controller  Created pod: instana-agent-5gww8
>   Normal  SuccessfulCreate  82s                daemonset-controller  Created pod: instana-agent-fs4bw
>   Normal  SuccessfulCreate  82s                daemonset-controller  Created pod: instana-agent-x246d
>   Normal  SuccessfulDelete  81s                daemonset-controller  Deleted pod: instana-agent-x246d
>   Normal  SuccessfulDelete  81s                daemonset-controller  Deleted pod: instana-agent-fs4bw
>   Normal  SuccessfulDelete  81s                daemonset-controller  Deleted pod: instana-agent-5gww8
>   Normal  SuccessfulCreate  81s                daemonset-controller  Created pod: instana-agent-4t6s2
>   Normal  SuccessfulCreate  81s                daemonset-controller  Created pod: instana-agent-lpczz
>   Normal  SuccessfulCreate  81s                daemonset-controller  Created pod: instana-agent-c22vw
>   Normal  SuccessfulDelete  77s                daemonset-controller  Deleted pod: instana-agent-4t6s2
>   Normal  SuccessfulDelete  77s                daemonset-controller  Deleted pod: instana-agent-lpczz
>   Normal  SuccessfulCreate  77s                daemonset-controller  Created pod: instana-agent-zlz27
>   Normal  SuccessfulCreate  75s                daemonset-controller  Created pod: instana-agent-tztpw
>   Normal  SuccessfulDelete  75s                daemonset-controller  Deleted pod: instana-agent-zlz27
>   Normal  SuccessfulDelete  75s                daemonset-controller  Deleted pod: instana-agent-tztpw
>   Normal  SuccessfulCreate  73s                daemonset-controller  Created pod: instana-agent-4k6mn
>   Normal  SuccessfulDelete  64s                daemonset-controller  Deleted pod: instana-agent-tn98q
>   Normal  SuccessfulDelete  49s                daemonset-controller  Deleted pod: instana-agent-6m8bc
>   Normal  SuccessfulDelete  26s (x2 over 36s)  daemonset-controller  (combined from similar events): Deleted pod: instana-agent-b4cxc
>   Normal  SuccessfulCreate  16s (x5 over 64s)  daemonset-controller  (combined from similar events): Created pod: instana-agent-kpcql
> ```
> 
> It is not happening on current main, so we can at least pin-point the problematic commit I think.

Nevermind, seemed to be a temporary issue, today I could not reproduce it any more. Maybe the testlab infra was too slow and the pods got killed :man_shrugging:

pkg/k8s/object/builders/agent/secrets/config.go
```diff
@@ -87,7 +87,7 @@ func (c *configBuilder) data() (map[string][]byte, error) {
 	if c.Spec.Agent.ConfigurationYaml != "" {
 		data["configuration.yaml"] = []byte(c.Spec.Agent.ConfigurationYaml)
 	}
-	if otlp := c.Spec.OpenTelemetry; !otlp.GrpcIsEnabled() || !otlp.HttpIsEnabled() {
+	if otlp := c.Spec.OpenTelemetry; *otlp.Enabled.Enabled {
```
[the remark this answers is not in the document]
konrad-ohms: Ah... I think I got it now... so it must be a change in behavior... before we basically checked, if there was any explicit opt-out. If not, we did not need to render the config file, because the agent would enable otel by default. With the current change, this must change as well, because we can change ports. If the ports changed, we must render the config as well.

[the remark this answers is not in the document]
konrad-ohms: So this is the new behavior, I compared what is rendered into the configmap via `oc get secret instana-agent-config -o yaml | yq -r '.data."configuration-opentelemetry.yaml"' | base64 -d`
- no grpc config -> that's fine
```yaml
com.instana.plugin.opentelemetry:
    enabled: true
    grpc:
        enabled: true
    http:
        enabled: true
```
**- spec.opentelemetry.enabled: false -> no configuration-opentelemetry.yaml created at all -> this is incorrect I think.**
- spec.opentelemetry.grpc.enabled: false
```yaml
com.instana.plugin.opentelemetry:
    enabled: true
    grpc:
        enabled: false
    http:
        enabled: true
```
- spec.opentelemetry.http.enabled: false
```
com.instana.plugin.opentelemetry:
    enabled: true
    grpc:
        enabled: true
    http:
        enabled: false
```
- spec.opentelemetry.http.enabled: false and port:4417
```yaml
com.instana.plugin.opentelemetry:
    enabled: true
    grpc:
        enabled: true
    http:
        enabled: false
        port: 4417
```
- only set both ports
```
com.instana.plugin.opentelemetry:
    enabled: true
    grpc:
        enabled: true
        port: 4418
    http:
        enabled: true
        port: 4417
```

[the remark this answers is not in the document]
FredrikAtIBM: Yes, I think you got my intention right. There are no getters, just default value that gets overridden by configuration input.

pkg/k8s/object/builders/common/ports/ports_builder_test.go
FredrikAtIBM: Yes I think this is something that I somehow missed

PR #276, opened 2025-06-27
Allow Kubernetes env vars on agent.pod.env

konrad-ohms:
## Why

As of now, the agent can only use a plain key-value map for agent.env. Kubernetes has more options, e.g. to expose a configmap or a secret as environment variable or even to access runtime generated data, like pod names.

## What

We do not want to break existing users, but offer a more advanced function by defining agent.pod.env which allow the entire Kubernetes spec for env vars v1.

## Checklist

- [x] Backwards compatible?


FelixMarxIBM (approved):
LGTM

PR #270, opened 2025-05-29
Remote agent operator

rorywelch:
## Why
Some sensor integrations that are installed through the agent should only be installed on a singular node as we do not want repeated copies of the same data being sent from the multiple agents to Instana. 

## What
Created a "Remote Agent" Operator which will monitor for new RemoteAgent CRs in the cluster. Each CR will deploy a singular unprivileged agent with the desired sensors configured. 

The Remote Agent operator follows the same logic as the existing Instana Agent operator except it is unprivileged within the cluster, and does not mount local host mount points.

The remote agent inherits values from the instana agent installed in the same namespace.

[Design and Soultion Doc](https://ibm.ent.box.com/notes/1823223338397)

## References

- [Story / Card](https://jsw.ibm.com/browse/INSTA-31092)
- [Documentation](https://ibm.ent.box.com/notes/1823223338397)

## Checklist

- [ ] [Release notes](https://github.ibm.com/instana/docs/blob/main/src/pages/releases/agent_operator_notes/index.md) in public docs updated?
- [ ] unit/e2e test coverage added or updated?

Note: Remember to run a [helm chart](https://github.ibm.com/instana/instana-agent-charts) release after the the operator release to make the changes available thru helm.

PR #278, opened 2025-07-07
Remote Operator V1

rorywelch:
## Why
Some sensor integrations that are installed through the agent should only be installed on a singular node as we do not want repeated copies of the same data being sent from the multiple agents to Instana. 

Pull Request based on https://github.com/instana/instana-agent-operator/pull/270, using a branch from inside the repository instead of a forked repository

## What
Created a "Remote Agent" Operator which will monitor for new RemoteAgent CRs in the cluster. Each CR will deploy a singular unprivileged agent with the desired sensors configured. 

The Remote Agent operator follows the same logic as the existing Instana Agent operator except it is unprivileged within the cluster, and does not mount local host mount points.

The remote agent inherits values from the instana agent installed in the same namespace.

[Design and Soultion Doc](https://ibm.ent.box.com/notes/1823223338397)

## References

- [Story / Card](https://jsw.ibm.com/browse/INSTA-31092)
- [Documentation](https://ibm.ent.box.com/notes/1823223338397)

## Checklist


konrad-ohms:
Note: PR https://github.com/instana/instana-agent-operator/pull/276 was merged since you started working on the remote agent. We would like to support that on the remote agent equally, so we would need to add the option agent.pod.env.

ChinmayGitHub:
#Reminder 
This is targeted for R301 (FF: 17th, GA: 28th)

rorywelch:
> For the hostname, I saw you derive the name from the remoteagent CR.
I am wondering, if we should allow an explicit overwrite, as we could technically deploy multiple agentremote CRs into different namespace with the same name.
Also it will not be a fully qualified domain name:

The hostname does also use the namespace in its creation as of now:
`hostname := fmt.Sprintf("instana-agent-r-%s-%s", d.GetNamespace(), d.getName())`  
which results in prefix: `instana-agent-r-`    namespace:` instana-agent-`    agent CR name: `remote-agent`

It does look a bit awkward and I can easily implement a manual override or default to the above logic (I will look to get the node dns somehow and use the agentCR name and namespace  as a prefix to the fully qualified domain name as a default?).


For `Also, can you change the remote_configuration_yaml in the CR to configuration_yaml? I think we should be as close to the original agent CR as possible.` I will remove the remote_configuration_yaml from the instanageagentremoteSpec, and just utilize the cofiguration_yaml value under `spec: agent: configuration_yaml: <value>` same as the instana agent

konrad-ohms:
Sounds good thanks.
Okay, you are right, I missed the namespace in the hostname. That being said, if you have multiple clusters deployed with the same namespace pattern, you will have multiple agents with the same hostname. Maybe have your logic as default but allow an overwrite via setting.

rorywelch:
Great thank you!

PR #284, opened 2025-07-17
ci: Add missing reslock timeout bump

konrad-ohms:
## Why

Forgot to increase the reslock timeout in https://github.com/instana/instana-agent-operator/pull/283

## What

Reslock ensures that we don't use shared resources at the same time between pipelines.

## Checklist

- [x] Backwards compatible?


konrad-ohms:
I killed to pipeline to not waste a CI run for it.

PR #290, opened 2025-08-06
Add 'agent.dependencyURLs' array and download listed URLs via initContainer to the agent's deployment folder

htreu:
hint: AI generated code

## Why

Instana agent might require external Java dependencies, which can not be served by IBM directly nor shipped with the agent installation. Those additional dependencies were previously downloaded via manual initContainer into the agent's container. The manual initContainer was part of the static agent YAML deployment. Since this is deprecated and about to be removed end of 2025, this PR introduces the replacement option for external Java dependencies.

## What

Provide a new array option "dependencyURLs" to list external jar files, which get downloaded into the Agent's `/opt/instana/agent/deploy` folder.

## References

- [Old Documentation (MySQL)](https://www.ibm.com/docs/en/instana-observability/1.0.301?topic=technologies-monitoring-mysql#mysql-version-800-and-later-support)

## Checklist

- [ ] Backwards compatible?

### DIFF
--- a/api/v1/inline_types.go
+++ b/api/v1/inline_types.go
@@ -164,6 +164,9 @@ type BaseAgentSpec struct {
 	// Sets the INSTANA_MVN_REPOSITORY_SHARED_PATH environment variable
 	// +kubebuilder:validation:Optional
 	MvnRepoSharedPath string `json:"instanaMvnRepoSharedPath,omitempty"`
+	// URLs to dependency files that will be fetched via an init container and shared with the agent pod
+	// +kubebuilder:validation:Optional
+	DependencyURLs []string `json:"dependencyURLs,omitempty"`
 	// Sets the AGENT_RELEASE_REPOSITORY_MIRROR_URL environment variable
 	// +kubebuilder:validation:Optional
 	MirrorReleaseRepoUrl string `json:"agentReleaseRepoMirrorUrl,omitempty"`
--- a/config/samples/instana_v1_instanaagent_with_dependency_urls.yaml
+++ b/config/samples/instana_v1_instanaagent_with_dependency_urls.yaml
@@ -1,0 +1,29 @@
+apiVersion: instana.io/v1
+kind: InstanaAgent
+metadata:
+  name: instana-agent
+  namespace: instana-agent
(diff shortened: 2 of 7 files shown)


konrad-ohms (changes requested):
Thanks for the feature update.
Could you open the PR on the main project instead of your fork? Otherwise CI is not running by default.

Also, there are a few changes required in existing files (see above) and an addition is necessary in an existing test:

```
$ git diff pkg/k8s/object/builders/agent/daemonset/daemonset_test.go
diff --git i/pkg/k8s/object/builders/agent/daemonset/daemonset_test.go w/pkg/k8s/object/builders/agent/daemonset/daemonset_test.go
index 5c3e2a8..1f36e05 100644
--- i/pkg/k8s/object/builders/agent/daemonset/daemonset_test.go
+++ w/pkg/k8s/object/builders/agent/daemonset/daemonset_test.go
@@ -344,6 +344,7 @@ func TestDaemonSetBuilder_getVolumes(t *testing.T) {
                gomock.Eq(volume.TlsVolume),
                gomock.Eq(volume.RepoVolume),
                gomock.Eq(volume.NamespacesDetailsVolume),
+               gomock.Eq(volume.DeployVolume),
        ).Return(expectedVolumes, expectedVolumeMounts)

        db := &daemonSetBuilder{
```

A context question.
Will it be sufficient to download artifacts in the /instana folder?
Your example will pull files like this

```
curl https://repo1.maven.org/maven2/mysql/mysql-connector-java/8.0.28/mysql-connector-java-8.0.28.jar --output /instana/mysql-connector-java-8.0.28.jar && curl https://repo1.maven.org/maven2/org/postgresql/postgresql/42.3.3/postgresql-42.3.3.jar --output /instana/postgresql-42.3.3.jar```

In the final agent pod, the files are available in `/opt/instana/agent/deploy`.
I am not fully sure, if sensors would be fine with that location or if that must be configurable/somewhere else.

Also, I guess this would be required for the remote agent deployment as well? See pkg/k8s/object/remote-agent.

But other than the remarks, the feature seem to work in general on my dev OCP cluster, thanks :+1:

PR #294, opened 2025-08-07
Upgrade end-to-end tests to use dotenv

FredrikAtIBM:
## Why

end-to-end tests had an extra step to source a dotenv-file before starting end-to-end tests.

## What

Add the ability to directly load a dotenv-file when starting end-to-end tests, update readme to reflect the latest change and make dotenv files ignored globally across the project.

## Checklist

- [x] Backwards compatible?


e2e/config.go
```diff
@@ -52,6 +53,11 @@ func init() {
 	var instanaApiKey, containerRegistryUser, containerRegistryPassword, containerRegistryHost, endpointHost, operatorImageName, operatorImageTag string
 	var found, fatal bool
 
+	err := godotenv.Load(".env")
+	if err != nil {
+		log.Warningln("Warning: error occurred while attempting to load the dotenv-file (expected to location: ./e2e/.env)", err)
```
konrad-ohms, line 58: Is the godotenv.Load using a relative path to the current location?
Or do we have a mismatch between where we load the file and what we print here.

We have `.env` in the code and `e2e/.env` in the message

konrad-ohms (approved):

e2e/config.go
FredrikAtIBM: The path is interpreted as relative to the current working directory. From the perspective of the repository, its ./e2e/.env, though as the `make`-target indicates the package to `github.com/instana/instana-agent-operator/e2e` the current working directory will be `./e2e/` from the perspective of the repository. Its a good point though, I'll update the warning to be a bit more mindful of this.

FredrikAtIBM: What about now?

PR #295, opened 2025-08-08
e2e: Add selective monitoring test for namespace labels

konrad-ohms:
## Why

Improve test coverage for selective monitoring on a namespace level.

## What

- Add a new test case with a real demo app to ensure we only monitor workload which we should.

- Added a Java demo application in e2e/java-demo-app/ with:
  - A simple HTTP server implementation
  - Dockerfile for containerization
  - Kubernetes deployment manifest

- Implemented a selective monitoring test that:
  - Deploys the Java app in three different namespaces with different labels
  - Configures the Instana agent with selective monitoring in opt-in mode
  - Verifies that only JVMs in opt-in namespaces are monitored

**Note**: Right now there is not a good way to see from the logs if a JVM got ignored correctly. For now the test is deploying all 3 apps at once, checks on which worker nodes they are running and than it checks all relevant agent logs.
The logs are searched for the INSTANA_TEST_POD_NAME env var which get's populated by the deployment itself. The name gets mapped back to the pod names found in a given namespace to decide, if we expect them to be ignored or to be monitored.
When a container is found, the PID is extracted, e.g. `adding new VM with PID 3537885` will pick `3537885`.
The PID can be used as an identifier to find an attachment log like `Performing initial attach to JVM with PID 3537885`.
Finally, a working attachment needs to be detected by searching for `Initial attach to JVM with PID 3537885 successful`.

The same happens for the other pod names, but if we find them, this means they were not correctly ignored and the test should fail.

## References

- [INSTA-10452](https://jsw.ibm.com/browse/INSTA-10452)

## Checklist

- [x] Backwards compatible?
- [ ] [Release notes](https://github.ibm.com/instana/docs/blob/main/src/pages/releases/agent_operator_notes/index.md) in public docs updated? n/a

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -6,7 +6,7 @@
 
 set -e
 
-echo "Running pre-commit hook...\n"
+echo "Running pre-commit hook..."
 
 # Get list of staged .go files
 FILES=$(git diff --cached --name-only --diff-filter=ACM | grep '\.go$' || true)
@@ -38,7 +38,7 @@ for FILE in $FILES; do
   git add "$FILE"
 done
 
-echo "✅ Linted successfully.\n"
+echo "✅ Linted successfully."
 
 # Run copyright check on all staged files (not only .go files)
 ALL_FILES=$(git diff --cached --name-only --diff-filter=ACM | grep -E '\.(go|sh|yml|yaml)$' || true)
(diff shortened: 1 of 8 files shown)


.githooks/pre-commit
```diff
@@ -6,7 +6,7 @@
 
 set -e
 
-echo "Running pre-commit hook...\n"
```
konrad-ohms, line 9: those \n strings were really shown on my terminal:

```
$ echo "Running pre-commit hook...\n"
Running pre-commit hook...\n
```

If we want to keep those, we might need to add -e to `echo`, but I don't think we need them here.

FelixMarxIBM (approved):

PR #503, opened 2026-04-30
fix: optimize controller memory consumption with namespace-scoped cache

cvrkota:
## Why

The watch was memory exhaustive in clusters with many resources.

## What

Configure the controller-runtime cache to only watch resources in the operator's namespace, significantly reducing memory footprint on clusters with many resources.

The cache now only stores resources from the operator's namespace (determined by POD_NAMESPACE environment variable) instead of caching all cluster-wide resources. ClusterRole and ClusterRoleBinding remain cached cluster-wide as they are cluster-scoped resources.

Additionally, added LabelChangedPredicate to the namespace watch to only trigger reconciliation when namespace labels change, preventing unnecessary reconciliations on namespace status updates.

## References

- [Story / Card](https://jsw.ibm.com/browse/INSTA-XXXXX) -- todo
- [CSP](https://ibmsf.lightning.force.com/lightning/r/Case/500gJ00000AhaQkQAJ/view)

## Checklist

- [x] Backwards compatible?

### DIFF
--- a/controllers/instanaagent_controller.go
+++ b/controllers/instanaagent_controller.go
@@ -29,11 +29,13 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	instanav1 "github.com/instana/instana-agent-operator/api/v1"
 	instanaclient "github.com/instana/instana-agent-operator/pkg/k8s/client"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -100,8 +101,18 @@ func main() {
...
+	// ClusterRole and ClusterRoleBinding are cluster-scoped so they will still be cached cluster-wide
+	log.Info("Configuring cache to watch only operator namespace", "namespace", operatorNamespace)
+
 	mgr, err := ctrl.NewManager(
 		cfg, ctrl.Options{
+			Cache: cache.Options{
+				DefaultNamespaces: map[string]cache.Config{
+					operatorNamespace: {},
+				},
+			},
```
FelixMarxIBM, line 115: This changes the cache to the operator namespace, right?

controllers/instanaagent_controller.go
```diff
@@ -77,6 +79,9 @@ func Add(mgr manager.Manager) error {
 
 				return requests
 			}),
+			// Only trigger reconciliation when namespace labels change
+			// This prevents unnecessary reconciliations on namespace status updates
+			builder.WithPredicates(predicate.LabelChangedPredicate{}),
```
FelixMarxIBM, line 84: This does not really change the requested entities, only the reconciliations?

cvrkota: Yes, we will still collect all namespaces on the startup, but we would only do reconciles if the label changes. This avoids unnecessary reconciles on the e.g. annotations/resource versions/etc. This is not a fix for mem consumption, but CPU load since we would have significantly lower number of reconcile loops

main.go
cvrkota: This one reduces the memory footprint and cache only resources in the operator namespace (for namespaced resources). The cluster-wide resources will as is - ClusterRole, ClusterRoleBinding and Namespace, but this is generally significantly lower number than pods/deployments/daemonset/etc

konrad-ohms:
Just rebased due to dependency changes on main, no need to wait for a new CI cycle, it was green before

PR #507, opened 2026-05-05
add namespace and pod readiness checks

cvrkota:
## Why

E2E tests are failing intermittently in the main pipeline with TestSelectiveMonitoring reporting "namespace instana-agent not found". This occurs when tests run after TestUpdateInstall, which triggers a full namespace deletion/recreation. The race condition happens because:

- Full reset deletes the entire instana-agent namespace (including operator deployment)
- Namespace is recreated but may still be in Terminating state
- Operator deployment check runs before namespace is fully ready
- Even when deployment exists, pods may not be ready to process CRs

## What

Added two safety checks to `SetupOperatorDevBuild()` in `e2e/agent_test_api.go`:

- Namespace readiness check - Waits up to 60s for namespace to be in Active phase (not Terminating) before proceeding
- Pod readiness verification - After deployment is ready, verifies operator pods are actually Running and Ready before creating CRs

## References

- NA

## Checklist

- [x] Backwards compatible?

### DIFF
--- a/e2e/agent_test_api.go
+++ b/e2e/agent_test_api.go
@@ -518,6 +518,12 @@ func SetupOperatorDevBuild() e2etypes.StepFunc {
 	return func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
 		ensureRegistrySecret(t, cfg)
 
+		// Ensure namespace is fully ready (not terminating) before proceeding
+		// This prevents race conditions after full namespace deletion/recreation
+		if err := waitForNamespaceReady(ctx, cfg, t); err != nil {
+			t.Fatalf("Namespace not ready: %v", err)
+		}
+
 		desiredImage := desiredDevBuildImage()
 		imageMatches, err := operatorImageMatches(ctx, cfg, desiredImage)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


konrad-ohms (approved):

konrad-ohms (commented):
I would prefer solving this with #508 rather than this PR.

The failure mode we are trying to address is a transient Deployment `NotFound` / delete-recreate window while waiting for readiness. The existing `WaitForDeploymentToBecomeReady()` fetches the Deployment before the readiness wait and then waits against that object, which makes the helper sensitive to stale state if the Deployment is recreated during install or reconciliation.

#508 fixes that behavior directly in the shared wait helper by refetching the Deployment throughout the wait and treating temporary `NotFound` as “not ready yet” until the timeout expires. That also improves all callers of `WaitForDeploymentToBecomeReady()`, not only the `SetupOperatorDevBuild()` path.

This PR adds useful-looking extra checks, but they are less targeted to the observed failure:
- the namespace readiness check runs after `ensureRegistrySecret()`, so it does not guard the earliest namespace-dependent operation
- pod readiness after Deployment availability does not address the stale Deployment / transient `NotFound` issue
- the change is scoped to `SetupOperatorDevBuild()`, while the fragile readiness behavior lives in the shared deployment wait helper

Maybe we close this in favor of #508, then revisit a smaller follow-up only, if we still see evidence that namespace Active state or operator pod readiness is a separate source of flakiness?

PR #554, opened 2026-07-27
fix: stop the k8sensor etcd-reader RBAC churn in kube-system

mansoorpervaiz:
## Why

Support case TS022558809: on AKS, a customer running the operator with the InstanaAgent CR in `base-mon` sees the `instana-agent-k8sensor-etcd-reader` Role and RoleBinding in `kube-system` get created, garbage collected, and recreated on every single reconcile, forever. Kubernetes does not support cross-namespace owner references for namespaced objects, so the ownerReference the operator stamps on those two objects (pointing at a CR in another namespace) is invalid and the garbage collector deletes them immediately. Their ArgoCD application is therefore permanently OutOfSync, and the apiserver takes a steady stream of pointless writes and deletes.

The Role is not needed in the first place. The k8sensor ClusterRole already grants `get/list/watch` on `services` and `endpoints` cluster-wide, and the ClusterRoleBinding binds the k8sensor ServiceAccount to it in every namespace, `kube-system` included. Etcd discovery itself is done by the operator using its own permissions (granted by the static Role in `config/rbac/role.yaml`), and the sensor only receives the resulting target URLs through `ETCD_TARGETS` and scrapes them over HTTP. The sensor never reads anything in `kube-system` through this Role.

## What

Two changes, one to fix the reported symptom and one to make the class of bug impossible:

1. Removed the `-etcd-reader` Role and RoleBinding builders and their wiring in `controllers/apply.go`. Verified before deleting that the permissions are fully covered by the existing k8sensor ClusterRole/ClusterRoleBinding, and that there is no install mode, flag, or spec field where the ClusterRole omits `services`/`endpoints` and this Role would be a needed least-privilege replacement. `Build()` on the ClusterRole builder is a static rule list wired unconditionally, and `-etcd-reader` appeared nowhere else in the repo (no bundle manifest, no e2e test, no chart).

   Existing installs clean themselves up two ways over, with no migration code. The two objects are recorded in the dependents ConfigMap, and `CleanupDependents` diffs every recorded entry against the current object set on every reconcile and deletes the difference explicitly by GVK/name/namespace, so they go on the first reconcile after the upgrade. Independently, any leftover that still carries the invalid ownerReference is collected by Kubernetes garbage collection once the operator stops recreating it, which is the same mechanism that produced the churn in the first place.

   Checked that `instana-agent-charts` does not template an `etcd-reader` Role of its own, so the operator-side fix is sufficient to clear the ArgoCD diff.

2. As defense in depth, `AddOwnerReference` now skips objects whose namespace differs from the CR's, so the operator can never again create an object that Kubernetes will immediately garbage collect. This does not affect cleanup: dependents are removed by the lifecycle manager, which deletes them explicitly by GVK/name/namespace (both on uninstall via `DeleteAll` and when an object goes out of scope), not by owner reference GC.

Deliberately out of scope: the separate `ETCD_TARGETS` env flapping issue is tracked on its own ticket and is untouched here.

### Testing

Both new tests were confirmed to fail before the fix and pass after:

- `TestApplyResourcesOnlyBuildsNamespacedObjectsInTheAgentNamespace` walks every builder `applyResources` hands to `ApplyAll` and asserts that no namespaced object is built outside the agent namespace. Before the fix it reports `Role instana-agent-k8sensor-etcd-reader is built outside of the agent namespace` plus the same for the RoleBinding.
- `TestTransformations_AddOwnerReference_OtherNamespace` covers same namespace, unset namespace, and different namespace. Before the fix the different-namespace case sets the invalid ref.

The namespace assertion test runs against a fully populated spec and a backend so the agent DaemonSet, the k8sensor Deployment, the PodDisruptionBudget and the secrets are actually emitted rather than skipped as empty Optionals, and it asserts exact per kind counts so the loop cannot quietly degrade to checking almost nothing when a builder drops out.

`go build ./...`, `go vet ./...` and `go test ./api/... ./controllers/... ./pkg/...` are clean, other than `TestInstanaAgentControllerTestSuite`, which fails identically on `main` in my environment because the envtest control plane binaries are not installed locally.

To confirm on a real non-`kube-system` install: with the CR in `base-mon`, `kubectl get role,rolebinding -n kube-system | grep etcd-reader` returns nothing after the first reconcile, the resourceVersion churn stops, and the ArgoCD app settles to Synced. `kubectl auth can-i list endpoints -n kube-system --as=system:serviceaccount:base-mon:instana-agent-k8sensor` still returns `yes` from the ClusterRoleBinding alone.

## References

- [Story / Card](https://jsw.ibm.com/browse/INSTA-105463)
- Support case TS022558809

## Checklist

- [x] Backwards compatible?


jon-rabe (changes requested):
## Review: `cleanupLegacyEtcdReaderRBAC` appears to be redundant

The core fix here is correct and well-reasoned — removing the builders and hardening `AddOwnerReference` is exactly the right approach. My concern is with `controllers/legacy_rbac_cleanup.go` and the associated `sync.Map` field on the reconciler.

### The lifecycle manager already handles this

The PR description justifies the manual cleanup with:

> *"the dependents ConfigMap only records them for the generation that created them"*

However, looking at `CleanupDependents` in `pkg/k8s/operator/lifecycle/dependent_lifecycle_manager.go` (line 125):

```go
for key, jsonString := range lifecycleConfigMap.Data {
```

It iterates over **every key in the ConfigMap**, not just the current generation's entry. The generation key is composed of `operatorVersion_crGeneration` (see `getCurrentGenKey`, line 211). After an operator upgrade the version component changes, so the old entry is retained under the previous key and is still diffed on every reconcile. Anything recorded in that old entry that is absent from the current builder list — including both `instana-agent-k8sensor-etcd-reader` objects — is **explicitly deleted by name** on the first reconcile after the upgrade, regardless of whether the objects still carry an ownerReference.

This is also exactly what `ApplyAll` in `operator_utils.go` does unconditionally on every reconcile (line 80):

```go
return o.dependentLifecycleManager.CleanupDependents(objects...)
```

So the lifecycle manager already covers the two scenarios cited in the PR description as motivating the manual cleanup:
- Objects that still carry the invalid ownerReference are deleted explicitly by the lifecycle manager (it deletes by GVK/name/namespace, not via GC).
- Objects that have already lost the ownerReference are also deleted explicitly for the same reason.

### The cost

`controllers/legacy_rbac_cleanup.go` adds 125 lines of production code, a `sync.Map` on the reconciler struct, and 251 lines of tests — for cleanup work the existing lifecycle manager performs as part of its normal operation on the first reconcile after upgrade.

### Request

Could you confirm whether there is a scenario where `CleanupDependents` would **not** catch the Role and RoleBinding after upgrading from a version that created them? If the lifecycle manager does cover this fully, I'd suggest removing `legacy_rbac_cleanup.go`, the `cleanupLegacyEtcdReaderRBACOnce` call in `applyResources`, and the `legacyEtcdReaderCleaned` field — keeping the two meaningful changes (builder deletion and the `AddOwnerReference` guard) without the added complexity.

mansoorpervaiz:
Thanks @jon-rabe , this is a good catch and I've removed it: `legacy_rbac_cleanup.go`, the call in `applyResources`, the `legacyEtcdReaderCleaned` field and the associated tests are all gone in e3c43cd, so the PR is back to just the builder deletion and the `AddOwnerReference` guard.

One thing worth correcting for the record though, because it changes the reasoning even though it doesn't change the outcome. You said:

> After an operator upgrade the version component changes, so the old entry is retained under the previous key

I don't think it does change. `transformations.GetVersion()` reads the `OPERATOR_VERSION` env var and falls back to `v0.0.1-dev`, and as far as I can tell nothing ever sets that var on the running pod. `config/manager/manager.yaml` only sets `POD_NAMESPACE`, and `OPERATOR_VERSION_PLACEHOLDER` is substituted into the `app.kubernetes.io/operator-version` label rather than into an env var (`manager.yaml:15,25`, `Makefile:346,358`). The Helm chart doesn't set it either. So the key stays `v0.0.1-dev_<generation>` across an upgrade, and what actually carries the pre-upgrade list is the `-dirty` snapshot at `dependent_lifecycle_manager.go:73-75`, not a distinct key. Am I reading that right, or is `OPERATOR_VERSION` injected somewhere I've missed?

That does leave a gap, which is what I was originally trying to cover: in `ApplyAll` the ConfigMap is written, then the objects are applied, and only then does `CleanupDependents` run. If that middle apply fails or the pod restarts in between, the next reconcile overwrites `-dirty` with the new list and the two objects are referenced by no key at all. Same outcome if the dependents ConfigMap is missing, since its name derives from the CR.

Having said that I think you're right to cut it. The leftovers still carry the invalid cross-namespace ownerReference, so garbage collection removes them anyway once the operator stops recreating them, which is the same mechanism that caused the churn. The manual delete only uniquely helps for an object that lost its ownerReference *and* landed in that window, which isn't worth 376 lines and a `sync.Map` on the reconciler.

jon-rabe (approved):
LGTM!

PR #555, opened 2026-07-27
fix: stop ETCD_TARGETS flapping between reconciles on vanilla Kubernetes

mansoorpervaiz:
## Why

On a K3s cluster with etcd metrics enabled, the k8sensor pod restart-loops for around an hour before it settles. A test engineer captured the Deployment flipping back and forth on alternating reconciles:

```
SpecUpdated  Env.6.Value: http://9.60.248.41:2381/metrics -> <nil>
```

Every flip is a new pod template hash, so every reconcile rolls the k8sensor pod. While that is happening the sensor is repeatedly killed and restarted, which means gaps in Kubernetes entity and control plane data for the cluster.

## What

`CreateDeploymentContext` discovers the etcd endpoints and compares them against the ones already on the Deployment. When they matched, it returned a `nil` deployment context as a "nothing to do" optimization.

That optimization does not hold, because the deployment builder does not patch the live object. It rebuilds the whole pod spec from scratch on every reconcile, and it only emits `ETCD_TARGETS` when the context is non-nil and carries targets. A `nil` context therefore renders a Deployment with no `ETCD_TARGETS` at all. Since the operator applies with server-side apply and `ForceOwnership`, and `Container.Env` is a map-typed list keyed on the env var name, the previously owned entry is not merged forward but removed. The next reconcile sees the env var missing, decides an update is needed, adds it back, and the cycle repeats.

The fix is to let the unchanged case fall through and build the same context as the changed case, so the discovered targets are always carried. The rendered Deployment is then byte identical between reconciles, the apply is a no-op, and the pod is only rolled when the discovered targets genuinely change. `compareAndUpdateETCDTargets` is now used only to choose the log message.

The same strip-and-roll cycle was reachable through two more doors, so those are closed here too. Both returned a nil context, with exactly the same downstream effect:

- **Discovery failing.** `ShouldSkipDiscovery` does a live `Exists` call to decide whether the cluster is OpenShift, and `BuildTargetsFromEndpoints` lists EndpointSlices. A single transient API error turned into "no targets", stripped `ETCD_TARGETS` and rolled the pod, and the next reconcile rolled it back. An error means the targets are unknown, not that there are none, so the targets already applied to the live Deployment are now retained.
- **No ready endpoints.** If the etcd service and its metrics port are still present but no endpoint is currently ready, discovery reported zero targets, which is indistinguishable from etcd having been removed. On a single node K3s cluster one unready pass was enough to roll the sensor. `DiscoveredETCDTargets` now carries an `Indeterminate` flag for that case and the applied targets are retained.

Retention is deliberately not unconditional, so the env var cannot get pinned forever: when discovery positively establishes that there is no etcd, meaning the service is gone, it has no metrics port, or discovery was skipped, the targets are still dropped and the k8sensor stops scraping them.

Two smaller things in the same area came out of review and are fixed here as well:

- `CheckCASecretExists` treated every error from its lookup as "the secret is not there", so one failed call dropped the CA mount and rolled the pod, exactly the same shape of bug as above. Only a clean not-found now means absent; anything else is reported and the applied settings are retained.
- A CR that configured its own `k8sSensor.etcd.ca` on a cluster where discovery also found a CA produced the CR volume and the discovered volume, both named `etcd-ca`, plus two `ETCD_CA_FILE` env vars. The API server rejects both as duplicates. The CR configuration now wins and the discovered CA is skipped, which matches how explicitly configured targets already take precedence.

`compareAndUpdateETCDTargets` was renamed to `etcdTargetsChanged`. It never updated anything, and now that it no longer gates the update either, the old name invited exactly the mistake this PR is fixing.

Finally, the ETCD discovery log lines that fire on every reconcile and say nothing new in steady state moved to `V(1)`. The inconclusive-discovery line stays at the default level, since that one is an anomaly worth seeing.

Scope notes:

- The OpenShift path is not affected. `setupOpenShiftETCDMonitoring` returns early and always hands back a non-nil context, so it never hits this skip.
- This is independent of the cross-namespace ownerReference issue on the etcd-reader Role and RoleBinding. Those builders only take the agent and never see the deployment context.

Test coverage:

- `TestETCDTargetsRemainStableAcrossReconciles` drives two consecutive reconciles and asserts the whole pod template is unchanged the second time round, since any difference at all changes the pod template hash and rolls the pod. Against the old code it fails with `ETCD_TARGETS` missing from the second render, which is the reported bug reproduced as a unit test.
- `TestETCDCASettingsRemainStableAcrossReconciles` and `TestETCDCASettingsRetainedWhenDiscoveryFails` do the same for the CA, which drives an env var, a volume and a volume mount off the same context and so flapped in lockstep with the targets.
- `TestETCDTargetsRetainedWhenDiscoveryFails` and `TestETCDTargetsRetainedWhenDiscoveryInconclusive` cover the two transient paths. Both fail against the old code with the env var stripped.
- `TestETCDTargetsClearedWhenETCDIsGone` and `TestETCDTargetsUpdateWhenDiscoveryChanges` are the guards in the other direction: retention must not pin the env var, and a real change in the discovered targets still propagates.
- `TestETCDTargetsNotRetainedWithoutExistingDeployment` covers a fresh install where discovery fails before anything has been applied.
- `TestDeploymentBuilder_CustomETCDCAWinsOverDiscoveredCA` pins the CA precedence. Against the old code it fails three times over, with two `ETCD_CA_FILE` env vars, two `etcd-ca` volumes and two `etcd-ca` mounts. `TestDeploymentBuilder_DiscoveredETCDCAUsedWithoutCustomCA` is its counterpart, confirming the discovered CA is still mounted when the CR configures none.
- `TestCheckCASecretExists` gained a case for a failing lookup, and `TestDiscoverETCDEndpoints` one for the error propagating.
- Two existing expectations changed with the behaviour: the "same targets" case in `TestCreateDeploymentContext_SimplifiedTests` asserted the nil context, and `TestDiscoverETCDEndpoints` asserted a nil result when no endpoints were ready.

## References

## Checklist

- [x] Backwards compatible?

### DIFF
--- a/controllers/apply.go
+++ b/controllers/apply.go
@@ -210,7 +210,9 @@ func etcdTargetsChanged(
 	discoveredTargets []string,
 	log logr.Logger,
 ) bool {
-	log.Info(
+	// These run on every reconcile and say nothing new in steady state, so they are
+	// kept off the default log level
+	log.V(1).Info(
 		"Comparing current ETCD targets with discovered targets to determine if update is needed",
 	)
 
@@ -228,10 +230,10 @@ func etcdTargetsChanged(
 		currentTargets = strings.Join(currentTargetsList, ",")
 	}
 
-	log.Info("Target comparison details",
+	log.V(1).Info("Target comparison details",
 		"currentTargets", currentTargets,
 		"newTargets", newTargets,
-		"needsUpdate", currentTargets != newTargets)
+		"targetsChanged", currentTargets != newTargets)
 
 	// Return true if targets are different (update needed)
 	return currentTargets != newTargets
@@ -576,13 +578,14 @@ func CreateDeploymentContext(
 		// deployment builder always rebuilds the full pod spec, so a nil context
 		// renders a Deployment without ETCD_TARGETS, the server-side apply strips
 		// the env var and rolls the pod, and the next reconcile adds it back again.
-		logger.Info("ETCD targets unchanged, keeping them on the Deployment")
+		// This is the steady state, so it stays off the default log level.
+		logger.V(1).Info("ETCD targets unchanged, keeping them on the Deployment")
 	}
 
 	// Use sorted targets for consistency
 	sortedTargets := getSortedTargets(discoveredETCD.Targets)
 
-	logger.Info("Using discovered ETCD targets", "targets", sortedTargets)
+	logger.V(1).Info("Using discovered ETCD targets", "targets", sortedTargets)
 	deploymentContext := &k8ssensordeployment.DeploymentContext{
 		DiscoveredETCDTargets: sortedTargets,
 	}
--- a/controllers/apply.go
+++ b/controllers/apply.go
@@ -503,13 +503,14 @@ func CreateDeploymentContext(
 
 	if err != nil {
 		logger.Info("K8sensor deployment not found, will create with discovered ETCD targets")
+	} else if compareAndUpdateETCDTargets(existingDeployment, discoveredETCD.Targets, logger) {
+		logger.Info("ETCD targets changed, Deployment will be updated")
 	} else {
-		// Compare current and discovered ETCD targets
-		needsUpdate := compareAndUpdateETCDTargets(existingDeployment, discoveredETCD.Targets, logger)
-		if !needsUpdate {
-			logger.Info("ETCD targets unchanged, skipping Deployment update")
-			return nil, nil
-		}
+		// The context still has to carry the targets when they are unchanged. The
+		// deployment builder always rebuilds the full pod spec, so a nil context
+		// renders a Deployment without ETCD_TARGETS, the server-side apply strips
+		// the env var and rolls the pod, and the next reconcile adds it back again.
+		logger.Info("ETCD targets unchanged, keeping them on the Deployment")
 	}
 
 	// Use sorted targets for consistency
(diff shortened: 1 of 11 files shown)


Ferenc- (changes requested):
As mentioned [on Slack](https://instana.slack.com/archives/C02AX69RE4U/p1785247448675309?thread_ts=1785235649.597909&cid=C02AX69RE4U).
Let's double check the necessity of `ETCD_TARGETS`.

mansoorpervaiz:
Closing this one. It was only ever kept open as a reference while we split it into smaller PRs, and everything in it has now shipped: #557 (a CR configured ETCD CA wins over the discovered one) and #561 (stop setting the ETCD env vars the k8sensor does not read) are both merged.

Worth recording why the final fix looks nothing like what is in here. This PR stabilised `ETCD_TARGETS` so it stopped flapping. Review then established that the k8sensor never reads `ETCD_TARGETS` at all, and has not since October 2025, so the correct fix was to stop setting it rather than to keep it steady. The same turned out to be true of `ETCD_METRICS_URL` and `ETCD_REQUEST_TIMEOUT`, and of the `k8sSensor.etcd.targets` field on the CR, which is now deprecated and ignored.

I checked the reported behaviour against current `main` rather than assuming it: three consecutive reconciles with identical cluster state render byte identical pod templates, so the pod is not rolled, and `compareAndUpdateETCDTargets`, the function this PR named as the root cause, no longer exists.

PR #556, opened 2026-07-28
fix: stop ETCD_TARGETS flapping when the discovered targets are unchanged

mansoorpervaiz:
## Why

On a K3s cluster with etcd metrics enabled, the k8sensor pod restart-loops for around an hour before it settles. A test engineer captured the Deployment flipping back and forth on alternating reconciles:

```
SpecUpdated  Env.6.Value: http://9.60.248.41:2381/metrics -> <nil>
```

Every flip is a new pod template hash, so every reconcile rolls the k8sensor pod. While that is happening the sensor is repeatedly killed and restarted, which means gaps in Kubernetes entity and control plane data for the cluster.

## What

`CreateDeploymentContext` discovers the etcd endpoints and compares them against the ones already on the Deployment. When they matched, it returned a `nil` deployment context as a "nothing to do" optimization.

That optimization does not hold, because the deployment builder does not patch the live object. It rebuilds the whole pod spec from scratch on every reconcile, and it only emits `ETCD_TARGETS` when the context is non-nil and carries targets. A `nil` context therefore renders a Deployment with no `ETCD_TARGETS` at all. Since the operator applies with server-side apply and `ForceOwnership`, and `Container.Env` is a map-typed list keyed on the env var name, the previously owned entry is not merged forward but removed. The next reconcile sees the env var missing, decides an update is needed, adds it back, and the cycle repeats.

The fix is to let the unchanged case fall through and build the same context as the changed case, so the discovered targets are always carried. The rendered Deployment is then byte identical between reconciles, the apply is a no-op, and the pod is only rolled when the discovered targets genuinely change.

The production change is nine lines in `controllers/apply.go`; the rest is test coverage.

Scope notes:

- The OpenShift path is not affected. `setupOpenShiftETCDMonitoring` returns early and always hands back a non-nil context, so it never hits this skip.
- This is independent of the cross-namespace ownerReference issue on the etcd-reader Role and RoleBinding. Those builders only take the agent and never see the deployment context.
- Review of this change surfaced several related problems in the same area: the targets are also dropped when discovery fails or returns no ready endpoints, `CheckCASecretExists` reports every lookup error as a clean absence, and a CR configured ETCD CA collides with a discovered one. Those are deliberately **not** fixed here, to keep this PR to the reported bug. 

Test coverage:

- `TestETCDTargetsRemainStableAcrossReconciles` drives two consecutive reconciles and asserts the container env is unchanged the second time round. Against the old code it fails with `ETCD_TARGETS` missing from the second render, which is the reported bug reproduced as a unit test.
- `TestETCDTargetsUpdateWhenDiscoveryChanges` confirms the env var is not pinned and a real change in the discovered targets still propagates.
- The existing "same targets" case in `TestCreateDeploymentContext_SimplifiedTests` asserted the nil context, so it was updated to expect the targets to be retained.

## References

## Checklist

- [x] Backwards compatible?

### DIFF
--- a/controllers/apply.go
+++ b/controllers/apply.go
@@ -503,13 +503,14 @@ func CreateDeploymentContext(
 
 	if err != nil {
 		logger.Info("K8sensor deployment not found, will create with discovered ETCD targets")
+	} else if compareAndUpdateETCDTargets(existingDeployment, discoveredETCD.Targets, logger) {
+		logger.Info("ETCD targets changed, Deployment will be updated")
 	} else {
-		// Compare current and discovered ETCD targets
-		needsUpdate := compareAndUpdateETCDTargets(existingDeployment, discoveredETCD.Targets, logger)
-		if !needsUpdate {
-			logger.Info("ETCD targets unchanged, skipping Deployment update")
-			return nil, nil
-		}
+		// The context still has to carry the targets when they are unchanged. The
+		// deployment builder always rebuilds the full pod spec, so a nil context
+		// renders a Deployment without ETCD_TARGETS, the server-side apply strips
+		// the env var and rolls the pod, and the next reconcile adds it back again.
+		logger.Info("ETCD targets unchanged, keeping them on the Deployment")
 	}
 
 	// Use sorted targets for consistency
--- a/controllers/apply_etcd_flap_test.go
+++ b/controllers/apply_etcd_flap_test.go
@@ -1,0 +1,172 @@
+/*
+(c) Copyright IBM Corp. 2025
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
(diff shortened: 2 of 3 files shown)


Ferenc- (changes requested):
As mentioned [on Slack](https://instana.slack.com/archives/C02AX69RE4U/p1785247448675309?thread_ts=1785235649.597909&cid=C02AX69RE4U).
Let's double check the necessity of `ETCD_TARGETS`.

mansoorpervaiz:
Good call on this one, `ETCD_TARGETS` really is unused.

It is declared as a constant in `net/etcd_client.go` and `plugins/control_plane/etcd.go` in k8sensor and read nowhere. It was live in the original ETCD metrics work (INSTA-55701), where `initETCDMetrics` went `ETCD_METRICS_URL`, then `ETCD_TARGETS`, then auto discovery. That read went away when discovery moved into `component_discovery.go` (INSTA-65515) and the constants were left behind. Today k8sensor finds the endpoints itself, from the `etcd-metrics` service or from pods labelled `component=etcd`.

While digging I found two more in the same state: `ETCD_METRICS_URL` is dead by the same history, and `ETCD_REQUEST_TIMEOUT` does not appear in the k8sensor source at all. So rather than stabilise values nothing reads, I have removed all three. The TLS settings (`ETCD_CA_FILE`, `ETCD_CERT_FILE`, `ETCD_KEY_FILE`) are genuinely read, so those stay along with the `etcd-ca` mount.

Removing the endpoints also kills the flap at its source. With no targets to compare there is nothing to check against the live Deployment, so the "targets unchanged, skip the update" path that returned a nil context and let the apply strip the env var is gone with it.

Superseded by #561, closing this one. Thanks for pushing back on it.

PR #557, opened 2026-07-28
fix: let a CR configured ETCD CA win over the discovered one

mansoorpervaiz:
## Why

A cluster that configures its own ETCD CA on the InstanaAgent CR, while ETCD discovery also finds a CA of its own, renders a k8sensor Deployment that the API server rejects.

The container ends up with two `ETCD_CA_FILE` environment variables, and the pod spec with two volumes and two volume mounts all named `etcd-ca`. Duplicate names in those lists are invalid, so the apply fails and the k8sensor is never updated.

The precondition is narrow, it needs `k8sSensor.etcd.ca` set on the CR, no explicit `k8sSensor.etcd.targets`, and an `etcd-ca` secret present in the agent namespace for discovery to pick up, but when it is met the component simply does not deploy.

## What

The deployment builder emits the ETCD CA from two independent places:

- from the CR, via `ETCDCAFileEnv` and `volume.ETCDCAVolume`, whenever `k8sSensor.etcd.ca.mountPath` or `.secretName` is set
- from discovery, via the deployment context, whenever an `etcd-ca` secret was found

Neither knew about the other, so when both applied, both were emitted.

The CR configuration now wins and the discovered CA is skipped, which is the same precedence the builder already applies to targets: discovered targets are only added when `k8sSensor.etcd.targets` is empty. Two guards, one on the env var and one on the volume, in `pkg/k8s/object/builders/k8s-sensor/deployment/deployment.go`.

This is pre-existing and independent of the `ETCD_TARGETS` flapping work. It was found while reviewing that change, and it applies cleanly on its own, so it is raised separately rather than bundled in.

Test coverage:

- `TestDeploymentBuilder_CustomETCDCAWinsOverDiscoveredCA` pins the precedence. Against the old code it fails three times over, reporting two `ETCD_CA_FILE` env vars, two `etcd-ca` volumes and two `etcd-ca` mounts, and it also asserts the discovered targets are still applied so only the CA is left to the CR.
- `TestDeploymentBuilder_DiscoveredETCDCAUsedWithoutCustomCA` is the counterpart, confirming the discovered CA is still mounted when the CR configures none.

## References

## Checklist

- [x] Backwards compatible?

### DIFF
--- a/pkg/k8s/object/builders/k8s-sensor/deployment/deployment_etcd_ca_conflict_test.go
+++ b/pkg/k8s/object/builders/k8s-sensor/deployment/deployment_etcd_ca_conflict_test.go
@@ -1,5 +1,5 @@
 /*
-(c) Copyright IBM Corp. 2025
+(c) Copyright IBM Corp. 2026
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/k8s/object/builders/k8s-sensor/deployment/deployment_etcd_ca_conflict_test.go
+++ b/pkg/k8s/object/builders/k8s-sensor/deployment/deployment_etcd_ca_conflict_test.go
@@ -1,0 +1,210 @@
+/*
+(c) Copyright IBM Corp. 2025
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


pkg/k8s/object/builders/k8s-sensor/deployment/deployment_etcd_ca_conflict_test.go
```diff
@@ -0,0 +1,210 @@
+/*
+(c) Copyright IBM Corp. 2025
```
Ferenc-:
```suggestion
(c) Copyright IBM Corp. 2026
```

Ferenc- (changes requested):

Ferenc- (approved):

PR #561, opened 2026-07-31
fix: stop setting the ETCD env vars the k8sensor does not read

mansoorpervaiz:
## Why

On a K3s cluster with etcd metrics enabled, the k8sensor pod restart-loops for around an hour before it settles. The `ETCD_TARGETS` env var was being added and then removed on alternating reconciles, and every flip rolls a new pod:

```
```

The first version of this PR stabilised the value so it stopped flapping. Review asked the better question: is `ETCD_TARGETS` used at all? It is not, so this PR removes it instead.

## What

The k8sensor does not read `ETCD_TARGETS`. It discovers the etcd endpoints itself, from the `etcd-metrics` service or from pods labelled `component=etcd`, falling back to a default endpoint. The env var is declared as a constant in `net/etcd_client.go` and `plugins/control_plane/etcd.go` and referenced nowhere else.

It was live once: the original etcd metrics work had `initETCDMetrics` with a precedence of `ETCD_METRICS_URL`, then `ETCD_TARGETS`, then auto-discovery. That read was dropped when discovery moved into `component_discovery.go`, and the constants were left behind. The operator kept setting a value nothing consumed.

The same is true of two more variables the operator sets on OpenShift:

- `ETCD_METRICS_URL`, dead by the same history
- `ETCD_REQUEST_TIMEOUT`, which does not appear in the k8sensor source at all

So this PR stops setting all three. What the k8sensor does read, and what this PR keeps, is the TLS configuration: `ETCD_CA_FILE`, `ETCD_CERT_FILE` and `ETCD_KEY_FILE`, along with the `etcd-ca` volume mounted at the path the k8sensor defaults to.

Removing the endpoints also removes the flap at its source. The deployment context no longer carries targets, so there is nothing to compare against the live Deployment, and the "targets unchanged, skip the update" path that returned a `nil` context, and so rendered a Deployment with the env var missing for the server side apply to strip, is gone with it. The context is now derived from discovery alone and comes out the same on every reconcile.

That leaves the `k8sSensor.etcd.targets` field on the CR, which fed the same dead env var and is therefore equally inert. It is now deprecated and ignored, following the same pattern the repo already uses for `k8sSensor.featureFlags.crdMonitoring`: the field is still accepted so nobody's CR breaks, setting it logs a deprecation warning, and it produces no env var.

Removing that also required a fix in `ShouldSkipDiscovery`. Targets on the CR used to short circuit ETCD discovery entirely, which was correct when they were passed through as `ETCD_TARGETS`. Discovery still has to run to find the CA, so leaving that in place would have silently dropped the CA mount for anyone still setting the deprecated field.

Notes:
- The operator defaults to `k8sensor:latest` with `PullAlways`, so nothing regresses unless someone pins an image from before the discovery rework.
- The CR configured ETCD CA now takes precedence over the discovered one, for the env var and the volume. Without that guard the restructured code could emit `ETCD_CA_FILE` twice and two volumes named `etcd-ca`, which the API server rejects. This overlaps with #557; whichever merges second will have a small conflict here that resolves to the same thing.

Test coverage:

- `TestDeploymentBuilder_DiscoveryDoesNotSetETCDTargets` asserts discovery no longer injects endpoints while the CA is still passed through.
- `TestDeploymentBuilder_OpenShiftOmitsUnreadETCDEnvVars` asserts the same for `ETCD_METRICS_URL` and `ETCD_REQUEST_TIMEOUT`, and that the TLS settings stay.
- `TestDeploymentBuilder_CRETCDTargetsIgnored` and `TestEnvBuilder_ETCDTargetsEnv` cover the deprecated CR field, and `TestDeploymentBuilder_CRETCDTargetsDoNotSuppressTheCA` guards the discovery interaction above.
- `TestDeploymentBuilder_ETCDEnvIsStableAcrossBuilds` and `Context does not depend on the live Deployment` are the regression guards for the reported restart loop: the same inputs render the same env, volumes and mounts, and the context never reads the live Deployment.

## References

## Checklist

- [x] Backwards compatible?


Ferenc- (changes requested):
In `pkg/k8s/object/builders/common/env/env_builder_etcd_test.go` line 161, we still seem to have some logic requiring the creation of `ETCD_TARGET`.
Would it make sense to remove that?

mansoorpervaiz:
Good spot, that was the loose end. Done in c664d1b and d9071e2.

That test was covering `k8sSensor.etcd.targets` on the CR, which fed the same dead env var, so it is just as inert as the discovered one. I have deprecated it following the pattern the repo already uses for `k8sSensor.featureFlags.crdMonitoring`: the field is still accepted so nobody's CR breaks on upgrade, setting it logs a deprecation warning, and it no longer produces an env var. The field doc on `ETCDSpec.Targets` is marked deprecated too, so it shows up that way in the generated CRD.

One thing that fell out of it and is worth a look. `ShouldSkipDiscovery` used to return early when targets were set on the CR, which was right back when they were passed straight through as `ETCD_TARGETS`. Discovery still has to run to find the CA though, so leaving that in would have silently dropped the CA mount for anyone still setting the deprecated field. I removed the short circuit and added `TestDeploymentBuilder_CRETCDTargetsDoNotSuppressTheCA` to pin it.

Worth knowing that `TestShouldSkipDiscovery` in `etcd_discovery_test.go` does not actually exercise `DefaultETCDDiscoverer.ShouldSkipDiscovery`. It reimplements the logic inside a mock and asserts against that, so it passed happily while the real behaviour changed underneath it. I kept the mock in step and flipped the CR targets case, but it is testing itself rather than the code. Happy to rewrite it against the real implementation if you think it is worth doing here rather than in a follow up.

Ferenc- (approved):
> Happy to rewrite it against the real implementation if you think it is worth doing here rather than in a follow up.

Yes please. Let's get rid of those in a follow up as soon as we can.

Also now there is one field in the yaml that customers might still use, let's consider informing the user about the deprecation.

mansoorpervaiz:
Thanks! Two things from your notes.

On the mock based tests, agreed, follow up it is. I have filed it rather than pulling it into this PR. Worth recording what we know so whoever picks it up does not have to rediscover it: `TestShouldSkipDiscovery` reimplements the logic inside `MockETCDDiscoverer` and asserts against that, so it never touches `DefaultETCDDiscoverer.ShouldSkipDiscovery`. I confirmed the consequence by reinstating the short circuit that d9071e2 removed and running the whole `./controllers/` package: nothing failed. So that fix currently ships with no coverage at all, and the same shape is worth checking on the other mock backed tests in that file while we are in there.

On the deprecated field, good catch, that was a real gap. The operator side already logs a warning when it is set and marks it deprecated in the CRD description, but the chart is where customers actually look and it was still advertising it. The README even had `--set k8s_sensor.etcd.targets[...]` in the ETCD example. Raised instana/instana-agent-charts#212 to mark it deprecated in `values.yaml`, the parameter table and the example, and to drop those `--set` lines.

One deliberate choice there: the chart still passes the value through to the CR. Stripping it in the template looked tempting but makes things worse, since the operator can only log the deprecation warning if the chart hands the field over.

PR #562, opened 2026-08-04
WIP: fix: keep the ETCD CA in place when discovery cannot determine the state

mansoorpervaiz:
> **Stacked on #561.** The base branch is `INSTA-105465-drop-unused-etcd-env`, so the diff shown here is only the two commits on top. Please merge #561 first, after which this retargets to `main` cleanly.

## Why

INSTA-105465 was reported as the k8sensor pod restart-looping because `ETCD_TARGETS` was added and removed on alternating reconciles. #561 fixes that at the source by removing the env var, since the k8sensor never read it.

The same strip-and-roll cycle is still reachable for the ETCD **CA**, which the k8sensor genuinely does read. Two paths in `CreateDeploymentContext` treat "we could not find out" as "there is nothing there":

- **Discovery returns an error.** `ShouldSkipDiscovery` does a live `Exists` call to decide whether the cluster is OpenShift, and `BuildTargetsFromEndpoints` lists EndpointSlices. One transient API error means the context comes back nil, which drops `ETCD_CA_FILE`, the `etcd-ca` volume and its mount, and rolls the k8sensor pod. The next reconcile succeeds and rolls it again.
- **The etcd-ca lookup fails.** `CheckCASecretExists` returned a plain bool, so any error, not just a clean not-found, was reported as "the secret is not there" with exactly the same result.

There is also a third, quieter case: if the etcd service and its metrics port are still present but no endpoint happens to be ready on that pass, discovery reported zero targets, which is indistinguishable from etcd having been removed. On a single node cluster one unready pass is enough.

## What

The rule is that an unknown state is not an absent one.

- `CheckCASecretExists` now returns `(bool, error)`. Only a clean `IsNotFound` means absent; anything else is propagated.
- `DiscoveredETCDTargets` gains an `Indeterminate` flag, set when the etcd service and metrics port exist but no endpoint is ready.
- On a discovery error or an inconclusive result, `CreateDeploymentContext` reads the CA already applied to the live Deployment and keeps it instead of clearing it.

Retention is deliberately not unconditional, so the CA cannot get pinned forever. When discovery positively establishes the state, meaning the service is gone, it has no metrics port, discovery was skipped, or the etcd-ca secret is genuinely absent, the CA is still dropped. Only the discovered CA is ever retained: a CA configured on the CR mounts at its own path and is rendered from the spec every reconcile, so it is never mistaken for a discovered one.

Test coverage:

- `TestETCDCARetainedWhenDiscoveryFails` and `TestETCDCARetainedWhenDiscoveryInconclusive` cover the two transient paths. Both fail against the current code with the CA dropped.
- `TestETCDCAClearedWhenPositivelyGone` is the guard in the other direction, and asserts that a positive result never even reads the live Deployment.
- `TestETCDCANotRetainedWithoutExistingDeployment` covers a fresh install where discovery fails before anything is applied.
- `TestETCDCANotRetainedForCRConfiguredCA` pins the CR versus discovered distinction.
- `TestCheckCASecretExists` gained a failing-lookup case, and `TestDiscoverETCDEndpoints` cases for the inconclusive result and for the CA error propagating.

## References

## Checklist

- [x] Backwards compatible?


mansoorpervaiz:
Closing this without merging.

To be clear about what that means: the bug is real and is still present in `main`. When ETCD discovery hits a transient failure, `CreateDeploymentContext` returns a nil context, which strips `ETCD_CA_FILE` and the `etcd-ca` volume and rolls the k8sensor pod, then rolls it again once discovery recovers. It is the same defect class as INSTA-105465 but with a different trigger, and it is not the behaviour that ticket described, which is why it was split out in the first place.

The reason for closing rather than merging is that the design should change shape before it lands. Once #561 removed `ETCD_TARGETS`, the deployment context carries a single field, and that field is fully determined by whether the `etcd-ca` secret exists. The endpoint enumeration and the `Indeterminate` flag in this PR are largely left over from when the operator still had to hand the endpoints over. Several reviewers arrived at that independently, and rebuilding it smaller is a better use of time than rebasing this and carrying the extra machinery.

The things that cost real effort to find are written up for the follow up work so they are not lost with this branch:

- the discovered `etcd-ca` volume must be mounted `Optional`, otherwise a secret deleted underneath a live Deployment wedges the next pod in `ContainerCreating` and the k8sensor never starts, which takes the whole Kubernetes entity ingest with it
- do not identify the discovered CA by matching the `ETCD_CA_FILE` value. `k8sSensor.etcd.ca.mountPath` can be set to the same path, producing a byte identical env var with no volume present
- anything that reads the live Deployment has to distinguish `IsNotFound` from other errors, otherwise it reproduces the same unknown means absent conflation one level further down
- retention has to re-check that the `etcd-ca` secret still exists, or a deleted secret stays referenced indefinitely

Tracked as follow up work rather than by an open PR, so nothing here implies work is in flight.

PR #564, opened 2026-08-14
fix(INSTA-103709): Stop cross-namespace resource copy

Ferenc-:
## Why

https://jsw.ibm.com/browse/INSTA-103709

## What

Stop copying ETCD resources if the target namespace is not the operator's namespace.

## References

- [INSTA-103709](https://jsw.ibm.com/browse/INSTA-103709)

## Checklist

- [ ] Backwards compatible?


controllers/apply.go
```diff
@@ -420,6 +421,23 @@ func CreateDeploymentContext(
 	discoverETCD ETCDDiscoverFunc,
 ) (*k8ssensordeployment.DeploymentContext, error) {
 	if isOpenShift {
+		var ok bool
+		var operatorNamespace string
+		if operatorNamespace, ok = os.LookupEnv("POD_NAMESPACE"); !ok || operatorNamespace == "" {
+			operatorNamespace = "instana-agent"
```
mansoorpervaiz: When `POD_NAMESPACE` is missing we fall back to the literal `instana-agent` and then carry on copying, so a security control ends up guessing where we are running. Do you think we could fail closed here instead, ie if we cannot determine our own namespace we skip the ETCD copy entirely and log loudly? Worst case for a misconfigured deployment is then no ETCD metrics, rather than the etcd client key landing in a namespace we never actually verified.

controllers/apply.go
```diff
@@ -420,6 +421,23 @@ func CreateDeploymentContext(
 	discoverETCD ETCDDiscoverFunc,
 ) (*k8ssensordeployment.DeploymentContext, error) {
 	if isOpenShift {
+		var ok bool
+		var operatorNamespace string
```
mansoorpervaiz: Non blocking nit: `main.go` already has `defaultOperatorNamespace = "instana-agent"` plus this exact POD_NAMESPACE resolution for leader election. Would it be worth pulling it into one small helper both call, so the default cannot drift between the two later on?

controllers/apply.go
```diff
@@ -420,6 +421,23 @@ func CreateDeploymentContext(
...
+		if operatorNamespace, ok = os.LookupEnv("POD_NAMESPACE"); !ok || operatorNamespace == "" {
+			operatorNamespace = "instana-agent"
+		}
+
+		if agent.Namespace != operatorNamespace {
```
mansoorpervaiz: How confident are we about the OLM install path here? Our CSV only declares `installModes: AllNamespaces` and there is no `operatorframework.io/suggested-namespace` annotation, so an OperatorHub install can put the operator in `openshift-operators` while the CR goes into `instana-agent` (which is what our own alm-examples uses). In that shape `agent.Namespace != operatorNamespace` and ETCD monitoring silently switches off for everyone installing that way, on the one distro this whole feature exists for. Have we tried it against a real OLM install?

Even if it turns out fine, is it worth surfacing the skip somewhere the user can see it, like the CR status or a Warning event? Today the only signal is an Info log in the operator pod, so from the user's side ETCD metrics just quietly disappear after an upgrade.

controllers/apply.go
```diff
@@ -420,6 +421,23 @@ func CreateDeploymentContext(
...
+			logger.Info(
+				"Skipping OpenShift ETCD monitoring: agent CR namespace differs from operator namespace. "+
+					"ETCD credentials are only copied when the agent runs in the operator's own namespace.",
+				"agentNamespace",
+				agent.Namespace,
+				"operatorNamespace",
+				operatorNamespace,
+			)
+			return nil, nil
```
mansoorpervaiz, line 444: One side effect of returning here is that we also skip `cleanupCopiedETCDResources`, which used to run inside `setupOpenShiftETCDMonitoring`. So on any cluster that already has a copied `etcd-client` Secret in a non operator namespace (whether from an actual exploit or just from someone running the CR elsewhere), the credential the report is about stays exactly where it is after the upgrade. Should we clean up on this path too, so the fix remediates rather than only prevents?

One wrinkle if we do: `cleanupCopiedETCDResources` deletes by name only, so calling it for an arbitrary namespace would happily delete a user's unrelated Secret called `etcd-client`. Maybe teach it to check our own `instana.io/copied-from` annotation first, and then it is safe to call from here?

controllers/apply.go
```diff
@@ -420,6 +421,23 @@ func CreateDeploymentContext(
...
+		}
 		return setupOpenShiftETCDMonitoring(ctx, c, agent, logger), nil
```
mansoorpervaiz, line 446: Non blocking, defence in depth thought: the check sits one level above the actual sink. Today `setupOpenShiftETCDMonitoring` has exactly one caller so this is airtight, but would it be worth also asserting the destination inside `copyETCDResourcesToNamespace` or `buildETCDResourceObjectMeta`? That way a future caller cannot reintroduce the same hole without tripping over it, and the assertion sits right next to the line the report points at.

controllers/apply.go
```diff
@@ -420,6 +421,23 @@ func CreateDeploymentContext(
 	discoverETCD ETCDDiscoverFunc,
 ) (*k8ssensordeployment.DeploymentContext, error) {
 	if isOpenShift {
```
mansoorpervaiz: Bigger picture, and definitely not for this PR, the underlying shape is that any tenant who can create an InstanaAgent CR gets a cluster privileged operator to act on their input. With the ETCD copy closed, that same CR still makes us create a cluster scoped ClusterRole and ClusterRoleBinding whose subject is a ServiceAccount in the CR's own namespace (`pkg/k8s/object/builders/k8s-sensor/rbac/clusterrolebinding.go`), and that role includes `use` on the `privileged` SCC plus cluster wide read of configmaps and pods/log. The names are `<cr-name>-k8sensor` and cluster scoped, so a CR called `instana-agent` in a tenant namespace also collides with the real install's binding.

Should we raise a follow up card for that? It also makes me wonder whether the durable fix is to stop reconciling CRs outside the operator namespace at all (cache scoping or the event filter), which would subsume this check rather than sitting alongside it.

controllers/apply_test.go
```diff
@@ -68,8 +68,9 @@ var _ operator_utils.OperatorUtils = (*mockOperatorUtils)(nil)
 func TestCreateDeploymentContext_SimplifiedTests(t *testing.T) {
 	agent := &instanav1.InstanaAgent{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "test-agent",
-			Namespace: "test-namespace",
+			Name: "test-agent",
+			// From now on, only those namespaces get auto ETCD discovery on OpenShift, that match the operator's namespace and `instana-agent` is the assumed default.
+			Namespace: "instana-agent",
```
mansoorpervaiz, line 76: Good that the new subtests pin POD_NAMESPACE explicitly. This first one now reads it from the ambient environment though, so it passes on a clean machine and fails anywhere POD_NAMESPACE happens to be set. `POD_NAMESPACE=openshift-operators go test ./controllers/ -run TestCreateDeploymentContext_SimplifiedTests` fails on the `OpenShift discovers and copies ETCD resources` subtest. Given the flaky CI we have been chasing on this card, how about a `t.Setenv("POD_NAMESPACE", "instana-agent")` at the top of the parent test so every subtest is deterministic no matter what the runner injects??

controllers/apply_test.go
```diff
@@ -269,6 +270,230 @@ func TestCreateDeploymentContext_SimplifiedTests(t *testing.T) {
...
+			)
+			mockClient.AssertExpectations(t)
+		},
+	)
+
+	t.Run(
+		"OpenShift namespace validation - POD_NAMESPACE not set uses default",
+		func(t *testing.T) {
+			// Ensure POD_NAMESPACE is not set
+			t.Setenv("POD_NAMESPACE", "")
```
mansoorpervaiz: Small nit: the comment says not set, but `t.Setenv(..., "")` sets it to empty, so the `!ok` branch of the `LookupEnv` check never actually runs. How about `t.Setenv("POD_NAMESPACE", "anything")` followed by `os.Unsetenv("POD_NAMESPACE")` so it is genuinely unset and still restored when the subtest ends? Same for the fourth subtest below.

docs/adr-openshift-etcd-resource-copying.md
```diff
@@ -71,6 +74,7 @@ The operator (not the k8sensor pod) performs this copying using its existing clu
 ✅ **Standard Kubernetes Pattern**: Resource copying is a common operator pattern
 ✅ **Automated Sync**: Operator handles certificate rotation automatically
 ✅ **Clean Lifecycle**: Resources are garbage-collected with the InstanaAgent CR
+✅ **Flexible Deployment**: Supports InstanaAgent CR in any namespace, not just `instana-agent`
```
mansoorpervaiz: Doc nit now that the note at the top says ETCD monitoring only works when the CR is in the operator's namespace: this pro still reads Supports InstanaAgent CR in any namespace, and the rejected alternative above says Would prevent users from deploying the agent in custom namespaces. Could we reword both so the ADR does not contradict its own new constraint for the next reader?

docs/control-plane-and-etcd-metrics.md
```diff
@@ -202,8 +205,8 @@ does not pass the endpoints in.
 
 - **No ETCD metrics appearing**: Ensure the ETCD service exists in `kube-system` namespace with proper labels
 - **TLS connection errors**: Verify CA certificates are properly mounted
-- **Timeout errors**: Check network connectivity between the k8sensor and the ETCD endpoints
-- **k8sensor pod fails with "configmap 'etcd-ca-bundle' not found"**: This occurs on OpenShift when ETCD monitoring resources are missing from the `openshift-etcd` namespace. The operator will log "OpenShift ETCD CA bundle not found, ETCD monitoring will be disabled" but the pod should start successfully without ETCD monitoring. If the pod fails to start, this indicates the source ETCD resources (`etcd-metrics-ca-bundle` ConfigMap and `etcd-metric-client` Secret) need to be restored in the `openshift-etcd` namespace, or you can create empty placeholder resources in the `instana-agent` namespace as a temporary workaround.
+- **Timeout errors**: Check network connectivity and consider adjusting `ETCD_REQUEST_TIMEOUT`
```
mansoorpervaiz: Non blocking, only because it caught my eye: this timeout line looks unrelated to the namespace fix. Happy to keep it if it is deliberate, just checking it did not ride along by accident. Also while we are in this file, the resource names a few lines up (`etcd-metrics-ca-bundle`, `etcd-metric-client`) do not match the constants we actually use (`etcd-ca-bundle`, `etcd-client`). Pre existing, but confusing for anyone following the doc to debug a missing Secret

controllers/apply.go
Ferenc-: > Worst case for a misconfigured deployment is then no ETCD metrics, rather than the etcd client key landing in a namespace we never actually verified.

In my assessment this is not quite what is happening, because specifying `POD_NAMESPACE` is not compulsory at all, and if it is not provided then the operator installs to the same default `instana-agent` anyway.

controllers/apply_test.go
Ferenc-: Very true, that they should run with a clear env.
I have changed it to run with a cleared env as while unlikely the default `instana-agent` might change at one point.

controllers/apply_test.go
Ferenc-: Done.

controllers/apply.go
Ferenc-: > pulling it into one small helper both call

Done

controllers/apply.go
Ferenc-: > Should we clean up on this path too, so the fix remediates rather than only prevents?

I don't think that the responsibilities of the operator should scope creep into remediations.
Furthermore considering that what we are doing here is already a functionality degradation,
I am already expecting customers to demand that they can at least manually copy these credentials
into their own namespace for usage, so I don't think we should actively prevent that.

controllers/apply.go
Ferenc-: Good point! I have added another check in `copyETCDResourcesToNamespace`.

main.go
```diff
@@ -142,16 +143,14 @@ func main() {
 
 	// Get the namespace where the operator is running for leader election
 	// Prefer POD_NAMESPACE environment variable, fallback to default
-	operatorNamespace := os.Getenv("POD_NAMESPACE")
-	if operatorNamespace == "" {
-		operatorNamespace = defaultOperatorNamespace
+	operatorNamespace := env.GetOperatorNamespace()
+	if operatorNamespace == defaultOperatorNamespace {
```
mansoorpervaiz: Pulling this into `env.GetOperatorNamespace` reads much better. I think the branch got inverted in the move though. `POD_NAMESPACE` is set from the downward API in `config/manager/manager.yaml` and in the published CSV, so on a normal install `operatorNamespace` is `instana-agent` precisely because the env var says so, and we now log `POD_NAMESPACE not set, using default namespace. Consider setting POD_NAMESPACE environment variable.` on every startup of every standard deployment. That is the sort of line support ends up chasing. Since `GetOperatorNamespace` already logs both cases itself, could this whole block just go away?

Two smaller things in the same spot. The `else` is now empty, and our `.golangci.yml` only enables govet, ineffassign, unused, misspell, exhaustive, errcheck and lll, so nothing in CI will flag it for you. And `defaultOperatorNamespace` still lives here at line 52 as well as unexported in `pkg/env`, so the drift risk I was worried about is still there, just moved. Could `main.go` drop its copy and lean on the helper alone?

pkg/env/env.go
```diff
@@ -19,14 +19,40 @@ package env
...
 		return os.Getenv(operatorVersionEnvVarName)
 	}
 	return fallbackOperatorVersion
 }
+
+// GetOperatorNamespace returns the namespace where the operator is running.
+// It reads from POD_NAMESPACE environment variable, falling back to "instana-agent" if not set.
+func GetOperatorNamespace() string {
+	if ns := os.Getenv(operatorNamespaceEnvVarName); ns != "" {
+		log.Info("Leader election namespace set from POD_NAMESPACE", "namespace", ns)
```
mansoorpervaiz: Now that this helper is on the reconcile path, the message is not quite true any more. `CreateDeploymentContext` and `copyETCDResourcesToNamespace` both call it, so on OpenShift we get two of these per reconcile, in a context that has nothing to do with leader election. From a run on this branch:

```
env.go:48] "Leader election namespace set from POD_NAMESPACE" logger="env" namespace="instana-agent"
env.go:48] "Leader election namespace set from POD_NAMESPACE" logger="env" namespace="instana-agent"
```

How about dropping the logging from the getter entirely and letting each caller say what it is doing, or at least moving it to `V(1)` with a neutral message? A plain getter that logs on every call gets noisy quickly in a busy cluster, and it is the kind of line people grep for when they are debugging leader election.

mansoorpervaiz (approved):
Thanks for the updates, this looks good and you were right about `POD_NAMESPACE`. Two minor logging comments below, neither blocking. Approving.

main.go
Ferenc-: > could this whole block just go away?

Now it's gone.

>  The else is now empty

True, dropped.

> defaultOperatorNamespace still lives here 

Removed.

pkg/env/env.go
Ferenc-: Removed the logging from the getter, only callers log now.

mansoorpervaiz:
LGTM... 🚢

PR #565, opened 2026-08-17
INSTA-105914 - fix: namespace-qualify cluster-scoped RBAC names to prevent multi-tenant collisions

Jefiya-MJ:
## What

ClusterRole and ClusterRoleBinding names for the agent and k8s-sensor components were previously derived solely from the CR name (e.g. `instana-agent`, `instana-agent-k8sensor`). 
Two tenants in different namespaces who both create a CR with the same nameshare ONE ClusterRoleBinding cluster-wide. Whichever CR reconciles last hijacks the binding to its own namespace's ServiceAccount: the other tenant's agent ServiceAccount silently loses its cluster RBAC.

These are cluster-scoped resources, so two InstanaAgent CRs in different namespaces sharing the same CR name would overwrite each other's RBAC objects.

This change introduces namespace-qualified names (e.g. `tenant-a-instana-agent`) when the CR namespace differs from the ServiceAccount name. The default single-tenant install(namespace=`instana-agent`, name=`instana-agent`) is unaffected — names are unchanged.

## Why

In multi-tenant deployments where multiple InstanaAgent CRs exist in different namespaces with the same CR name, the operator was silently overwriting cluster-scoped RBAC objects across tenants, causing one tenant's RBAC to clobber another's.

## Changes

- `helpers.go`: Added `ClusterScopedRBACName()` and `ClusterScopedK8sSensorRBACName()` —
  return namespace-prefixed names when `namespace != ServiceAccountName()`, bare names otherwise
- `agent/rbac`, `k8s-sensor/rbac`: Switch ClusterRole and ClusterRoleBinding builders to use
  the new scoped name methods
- `main.go`: Added `cleanupLegacyAgentRBAC()` — runs at startup to delete orphaned bare-name
  RBAC objects left behind on upgrade; skips the default install where names are unchanged;
  only deletes objects owned by this operator (`managed-by` label check)
- `helpers_mock.go`: Generated mock updated for new interface methods
- Tests updated throughout

## References

- [INSTA-105914](https://jsw.ibm.com/browse/INSTA-105914)

## Checklist

- [ ] Backwards compatible?


main.go
```diff
@@ -321,6 +340,98 @@ func cleanupOldOperator(k8sClient k8sClient.Client) {
...
+				clusterRole:        saName,
+				clusterRoleBinding: saName,
+			},
+			// k8s-sensor component
+			oldRBACNames{
+				clusterRole:        saName + "-k8sensor",
+				clusterRoleBinding: saName + "-k8sensor",
+			},
+		)
+	}
```
konrad-ohms: Hi @Jefiya-MJ,
thanks for your fixes. They should work in general, but there is a small edge case which we could improve I think.

The candidates are not deduplicated.

The current code builds a `candidates` slice by iterating over all `InstanaAgent` CRs. Each CR
contributes two entries keyed on `saName` (agent) and `saName + "-k8sensor"` (k8s-sensor). When two CRs in **different namespaces** share the same CR name like `instana-agent` in both namespaces `instana-agent-a` and `instana-agent-b`, the  `ServiceAccountName()` returns `"instana-agent"` for both, so the slice ends up with the entry `"instana-agent"` **twice**:

```
instana-agent-a/instana-agent  →  saName = "instana-agent"  →  append "instana-agent"
instana-agent-b/instana-agent  →  saName = "instana-agent"  →  append "instana-agent"  ← duplicate
```

The loop then attempts `Get` + `Delete` on `"instana-agent"` a second time. This is not a
correctness bug as the `!errors.IsNotFound` guard on the `Get` silently swallows the second
attempt, but it is unnecessary work and makes the code misleading.

The logging is not entirely correct for that case.
```go
if err := k8sClient.Delete(ctx, oldCR); err != nil && !errors.IsNotFound(err) {
    log.Error(err, "Failed to delete legacy ClusterRole", "name", names.clusterRole)
} else {
    log.Info("Successfully deleted legacy ClusterRole", "name", names.clusterRole) // ← fires on NotFound too
}
```

The `else` branch fires whenever `err == nil` **or** `errors.IsNotFound(err)` is true. In the
duplicate-name scenario above, the second iteration reaches this branch with `NotFound` and logs `"Successfully deleted"` which is not true here.

### Fix proposal

Replace the `candidates` slice with a `map[string]struct{}` set to deduplicate names before the delete loop. Because each name is then visited exactly once, the object retrieved by `Get` is guaranteed to exist when `Delete` is called, so the `!errors.IsNotFound` guard on delete can also be dropped.

```go
// Use a set to deduplicate bare names: multiple CRs in different namespaces
// can share the same ServiceAccount name, producing identical old RBAC names.
seen := make(map[string]struct{})
for _, agent := range agentList.Items {
    h := helpers.NewHelpers(&agent)
    saName := h.ServiceAccountName()

    if agent.Namespace == saName {
        continue
    }

    seen[saName] = struct{}{}
    seen[saName+"-k8sensor"] = struct{}{}
}

for name := range seen {
    // --- ClusterRole ---
    oldCR := &rbacv1.ClusterRole{}
    if err := k8sClient.Get(ctx, types.NamespacedName{Name: name}, oldCR); err != nil {
        if !errors.IsNotFound(err) {
            log.Error(err, "Failed to get legacy ClusterRole", "name", name)
        }
    } else if oldCR.Labels[managedByLabel] != managedByValue {
        log.Info("ClusterRole found but not managed by this operator, skipping", "name", name)
    } else {
        log.Info("Deleting legacy ClusterRole", "name", name)
        if err := k8sClient.Delete(ctx, oldCR); err != nil {
            log.Error(err, "Failed to delete legacy ClusterRole", "name", name)
        } else {
            log.Info("Successfully deleted legacy ClusterRole", "name", name)
        }
    }

    // --- ClusterRoleBinding ---
    oldCRB := &rbacv1.ClusterRoleBinding{}
    if err := k8sClient.Get(ctx, types.NamespacedName{Name: name}, oldCRB); err != nil {
        if !errors.IsNotFound(err) {
            log.Error(err, "Failed to get legacy ClusterRoleBinding", "name", name)
        }
    } else if oldCRB.Labels[managedByLabel] != managedByValue {
        log.Info("ClusterRoleBinding found but not managed by this operator, skipping", "name", name)
    } else {
        log.Info("Deleting legacy ClusterRoleBinding", "name", name)
        if err := k8sClient.Delete(ctx, oldCRB); err != nil {
            log.Error(err, "Failed to delete legacy ClusterRoleBinding", "name", name)
        } else {
            log.Info("Successfully deleted legacy ClusterRoleBinding", "name", name)
        }
    }
}
```

konrad-ohms (changes requested):

main.go
Jefiya-MJ: Thanks, Konrad. Fixed the deduplication and logging issue.

konrad-ohms (approved):

PR #577, opened 2026-08-25
INSTA-108871: add RunAsNonRoot and RunAsUser to k8sensor deployment

jon-rabe:
## Why

The k8sensor pod fails to start on OCP 4.17+ when the `nonroot-v2` SCC is assigned to the k8sensor ServiceAccount. The `nonroot-v2` SCC enforces `runAsNonRoot: true`, which requires the kubelet to verify that the container will not run as root. On OCP 4.17+, this verification requires a numeric runAsUser to be present in the container SecurityContext. The kubelet no longer resolves symbolic usernames from the images `/etc/passwd` at admission time. The k8sensor image defines its user as the symbolic name k8sensor:k8sensor and the operator did not set runAsUser, causing the kubelet to reject the pod with:

> container has runAsNonRoot and image has non-numeric user (k8sensor), cannot verify user is non-root

## What

Added container-level SecurityContext to the k8sensor Deployment with
- RunAsUser: 1000
- RunAsNonRoot: true

## References

- [Story / Card](https://jsw.ibm.com/browse/INSTA-108871)
- [CSP](https://ibmsf.lightning.force.com/lightning/r/Case/500gJ00000FKcLRQA1/view)

## Checklist

- [x] Backwards compatible?


pkg/k8s/object/builders/k8s-sensor/deployment/deployment.go
```diff
@@ -313,6 +313,14 @@ func (d *deploymentBuilder) build() *appsv1.Deployment {
...
+							// is non-root when nonroot-v2 SCC injects runAsNonRoot:true. Without a
+							// numeric UID the kubelet rejects the pod because it cannot resolve the
+							// symbolic username "k8sensor" from /etc/passwd at admission time.
+							SecurityContext: &corev1.SecurityContext{
+								RunAsUser:    pointer.To(int64(1000)),
+								RunAsNonRoot: pointer.To(true),
+							},
 						},
 					},
 					// k8sensor is run as a "k8sensor" user (i.e: uid 1000), and thus reading the files from the secret volume
```
mansoorpervaiz: Non blocking, a few small thoughts now that this area has grown.

We've got 1000 in two places that have to agree, the `FSGroup` here and the `RunAsUser` you added above. How about pulling it into a named constant, something like `k8sensorUID = int64(1000)`, and using it for both? Then the coupling to the `k8sensor` user in the image is stated once instead of living as two independent magic numbers that someone could later change separately

Also, would it read better to set `RunAsUser` and `RunAsNonRoot` on this pod level `SecurityContext` rather than on the container? The comment right here already tells the uid 1000 story, so it would keep the whole explanation in one place, and pod level would automatically cover anything we add later like an init container or a sidecar. Genuinely happy either way if you deliberately wanted it scoped to the one container.

Last one, and it's really a follow up for a different repo rather than anything for this PR. Neither side actually states 1000 anywhere. In k8sensor's `Dockerfile.release` the build stage runs `useradd --gid k8sensor --shell /bin/false k8sensor` with no `--uid`, so 1000 is just the first free UID above `UID_MIN` in the `golang:1.26` base. It's 1000 today and we're now hardcoding that here, but if that base ever ships a user at 1000 the sensor quietly becomes 1001 and we'd be running it as a UID with no `/etc/passwd` entry. Low impact, since the binary is world executable and `FSGroup` covers the secret reads, so it would only bite something doing a user lookup. Still, do you think it's worth a one line follow up on k8sensor to pin `groupadd --gid 1000` and `useradd --uid 1000`, so the number is written down on the side that owns it?

pkg/k8s/object/builders/k8s-sensor/deployment/deployment_test.go
```diff
@@ -1104,3 +1105,26 @@ func TestValidatePollRateAcceptsSeconds(t *testing.T) {
...
+	builder.VolumeBuilder.(*MockVolumeBuilder).On("Build", mock.Anything).
+		Return([]corev1.Volume{{Name: "config"}}, []corev1.VolumeMount{{Name: "config"}})
+	builder.statusManager.(*MockStatusManager).On("SetK8sSensorDeployment", mock.Anything).Return()
+
+	container := builder.build().Spec.Template.Spec.Containers[0]
+
+	require.NotNil(t, container.SecurityContext)
+	require.NotNil(t, container.SecurityContext.RunAsUser,
+		"RunAsUser must be set: OCP 4.17+ kubelet cannot verify non-root from a symbolic username")
+	assert.NotZero(t, *container.SecurityContext.RunAsUser,
```
mansoorpervaiz, line 1126: Nit, non blocking. Since the whole point is that this UID has to match the `k8sensor` user baked into the image, would it be worth asserting the exact value, `assert.Equal(t, int64(1000), *container.SecurityContext.RunAsUser)`, rather than just non zero? `NotZero` would still pass if someone later changed it to 65532, which satisfies `nonroot-v2` but isn't the user the image was actually built around, and it wouldn't line up with the `FSGroup: 1000` below either.

Given the operator is the only way k8sensor gets installed, this assertion is really the only place that contract is written down, so I'd quite like it to name the number

mansoorpervaiz (approved):

pkg/k8s/object/builders/k8s-sensor/deployment/deployment.go
jon-rabe: Good point on using a constant for the k8sensor UID magic number - done.

Concerning the SecurityContext placement - the intent is deliberately to scope it to this one container. If we add an init container or sidecar later that needs different settings, inheriting from pod level could be surprising. Happy to revisit if this changes.

On the k8sensor dockerfile: I completely agree and also wondered about this. Worth noting: The implicit assumption on UID 1000 is not something introduced in this PR but has been there with the `FSGroup: 1000` before this change; this PR just makes it visible by adding a second consumer of that assumption. I will open a follow-up ticket that makes sure using 1000 for the k8sensor UID in the k8sensor image is not a coincidence but an intentional contract.

pkg/k8s/object/builders/k8s-sensor/deployment/deployment_test.go
jon-rabe: I would push back here. This test is explicitly about ensuring compatibility with the `nonroot-v2` SCC which does not require the UID to be exactly 1000. Asserting it in this test may produce difficult to understand test failures.

If we want to verify the 1000 explictly this should be its own test - however I would think going with the constant in the change above is probably sufficient. If its ever changed that would be an intentional change and having a test case that needs to be changed would only be a blocker.

Therefore I would prefer to leave it as is.

PR #594, opened 2026-09-08
ci: increase e2e timeout

cvrkota:
## Why

* we are hitting timeouts in the release pipeline

## What

* increase the timeout

## References

- NA
## Checklist

- [ ] Backwards compatible?


konrad-ohms (changes requested):
Please increase the reslock ttl as well, otherwise locks would be potentially released too early

Jefiya-MJ (approved):

